PR #6, opened 2018-09-19
Foundation CSS

IgnacioRodrigues:


foundation/foundation.css
```diff
@@ -0,0 +1,133 @@
...
+
+h3 {
+    font-size: 18px;
+    line-height: 22px;
+    color: #333333;
+    padding-bottom: 6px;
+}
+
+p {
+	  color: ;
```
mvillaseco: empty color and missaligned

foundation/foundation.css
```diff
@@ -0,0 +1,133 @@
...
+	  font-size: 15px;
+    line-height: 22px;
+    color: #666666;
+    padding-bottom: 6px;
+}
+
+small {
+	  font-size: 13px;
```
mvillaseco: missaligned

foundation/foundation.css
```diff
@@ -0,0 +1,133 @@
...
+    text-transform: uppercase;
+}
+
+/* BACKGROUNDS COLORS - grey */
+.darkgrey {
+    background-color: #333333; /* text */
+}
+
+.grey {
+	  background-color: #666666; /* icons fonts */
```
cbernat: Please check alignment through all the css, maybe you have an issue with your code editor.

foundation/foundation.css
```diff
@@ -0,0 +1,133 @@
+body {
+    font-family: Arial,Helvetica,Verdana,sans-serif;
+    font-size: 15px;
+    line-height: 22px;
+    color: #666666;
```
mvillaseco: I know that we only save 300 bytes if we use #666 instead of #666666;, so i'm not so sure if we should change this or not, i let someone else says something.

foundation/foundation.css
```diff
@@ -0,0 +1,133 @@
...
+
+.micro {
+    font-size: 11px;
+    line-height: 13px;
+    color: #666666;
+    text-transform: uppercase;
+}
+
+/* BACKGROUNDS COLORS - grey */
+.darkgrey {
```
mvillaseco: hmmm, maybe dark-grey?

foundation/grid.css
```diff
@@ -0,0 +1,22 @@
+/* GRID */
+.grid {
+    display: grid;
+    grid-gap: 20px;
+    grid-template-columns: repeat(auto-fit, minmax(50px, 1fr));
```
mvillaseco: 👍

mvillaseco: I know this is a foundation, but i don't like to have in the same class that declares grid a column template set.

How about adding a class especially for that column alignment.

foundation/layout.css
```diff
@@ -0,0 +1,47 @@
+main {
+    padding: 0 5%;
+}
+
+section {
+	width: 100%;
+	max-width: 1366px;
+	margin: 0 auto;
+	padding: 80px 0;
+    border-bottom: 1px solid #CCCCCC;
```
mvillaseco: missaligned

foundation/special.css
```diff
@@ -0,0 +1,82 @@
...
+
+small {
+	  font-size: 13px;
+    line-height: 18px;
+    margin-bottom: 3px;
+}
+
+/* SPACES */
+
+.margin--0 {
```
mvillaseco: double dash?

foundation/special.css
```diff
@@ -0,0 +1,82 @@
...
+	margin-: 0; /* when p is a title of a paragraph */
```
mvillaseco: extra dash here jajaj :)

foundation/foundation.css
```diff
@@ -0,0 +1,133 @@
...
+    line-height: 36px;
+    color: #333333;
+    padding-bottom: 18px;
+}
+
+h2 {
+    font-size: 24px;
+    line-height: 22px;
+    color: #333333;
+    padding-bottom: 14px;
```
mvillaseco: Here you use padding, but in special stylesheet you use margin :S.

foundation/foundation.css
mvillaseco: misaligned

foundation/foundation.css
```diff
@@ -0,0 +1,133 @@
...
+.darkgrey {
+    background-color: #333333; /* text */
+}
+
+.grey {
+	  background-color: #666666; /* icons fonts */
+}
+
+.lightgrey {
+	  background-color: #999999; /* placeholders, hovers*/
```
mvillaseco: same here

foundation/foundation.css
```diff
@@ -0,0 +1,133 @@
...
+.sylver {
+    background-color: #CCCCCC; /* strokes */
+}
+
+.snow {
+    background-color: #EAEAEA; /* disabled inputs */
+}
+
+.ghostwhite {
+	background-color: #F2F2F2; /* backgrounds grid, cebra */
```
mvillaseco: :)

foundation/foundation.css
MarianaMS: the color is in the code, looks like a bug from the code editor

foundation/foundation.css
MarianaMS: Some browsers present interpetation errors with # of only 3 numbers

foundation/foundation.css
MarianaMS: watch out! is true alignments are not declared in the headings, but doesn't seem right to me declare alignments in color classes

foundation/grid.css
MarianaMS: The grid isn't part of the foundation, I only use it to build the front layout

foundation/foundation.css
MarianaMS: I think that's why I called it special, I had to use margin because the function is separate the color background of the boxes ;)... not the same end

foundation/special.css
MarianaMS: This is a recommendation? I didn't write this with double dash =(

IgnacioRodrigues:
In this branch we made the class arrangements, css alignment spaces for the foundation. https://github.com/FromDoppler/Doppler-UI-System/pull/9 Please look at the Pull Request again! 🙏🏻

PR #24, opened 2019-02-14
fix(gulpfile): remove unused code and fix dist task

cbernat:
- Add sass:dist task
- Fix config to point css to dist/css as it points in templates/html index.html file


gulpfile.js
```diff
@@ -59,7 +53,7 @@ var config = {
   },
   folderDist: {
     base: 'dist',
-    css: 'dist',
+    css: 'dist/css',
```
andresmoschini: Is it fine? we do not want css files in root path?

gulpfile.js
```diff
@@ -115,6 +96,22 @@ gulp.task('sass:build', ['webfont'], function() {
     .pipe(gulp.dest(config.folderDev.css));
 });
 
+gulp.task('sass:dist', ['webfont'], function() {
```
andresmoschini: This code is almost the same as _sass_ task, did you consider reuse the code?

I am not so familiar with gulp tasks, maybe it is better having clear and duplicated code than ugly reusings.

gulpfile.js
```diff
@@ -31,13 +31,7 @@ var gulp = require('gulp'),
   mqpacker = require('css-mqpacker'),
   mqsorter = require('sort-css-media-queries'),
   // Image optimization plugin
-  imagemin = require('gulp-imagemin'),
-  // Concat plugin
-  concat = require('gulp-concat'),
-  // uglify plugin
-  uglify = require('gulp-uglify'),
```
andresmoschini: Is it possible also removing npm dependencies?

andresmoschini:
Great work, thanks!

gulpfile.js
cbernat: I think either way is the same. I had an issue with the html template being used for dist and for dev at the same time, the route in the html is hardcoded. Perhaps we don't need to have an html in the dist folder.

andresmoschini:
Jenkins CI is not ready yet, so, merge it whenever you want without paying attention to this message: 

image

gulpfile.js
andresmoschini: @cbernat clarified me that it was different only for dist, so she is making it consistent. I understand now. Thanks!

gulpfile.js
cbernat: Yes sure. Good catch!

GusBaamonde: No vamos a utilizar un minimo archivo custom.js como por ejemplo para interacciones del menu? O solo se utilizara puro css?

andresmoschini: En principio pensabamos intentaremos dejar afuera JS, cuando sea necesario se agregará.

¿Les parece bien?

GusBaamonde: Ok

gulpfile.js
cbernat: I researched this, and is better this way for now. I did not find another way around.

PR #25, opened 2019-02-15
Added new files to build docker and publish in the cdn

IgnacioRodrigues:
chore(build): added  files to build docker, modify gulpfile.js and modify package.json


gulpfile.js
```diff
@@ -93,22 +93,6 @@ gulp.task('sass:build', ['webfont'], function() {
     }))
```
lucianaDilizio: Take care, if I'm not wrong you're undoing the changes that @cbernat did.

build-w-docker.sh
```diff
@@ -0,0 +1,25 @@
...
+export MSYS2_ARG_CONV_EXCL="*"
+
+rm -rf dev
+rm -rf dist
+
+docker build --pull -t doppler-ui-system-source .
+docker run --rm \
+    -v `pwd`/dist:/work/dist \
+    -p 3500:3500 \
+    -e CHOKIDAR_USEPOLLING=true \
```
cbernat: why are you using this `CHOKIDAR_USEPOLLING=true \` in the build script? As I undestand this is used to poll for watch purposes.

gulpfile.js
cbernat: Yes @IgnacioRodrigues  please do not remove this, else dist task will cease to work again.

gulpfile.js
```diff
@@ -257,4 +240,4 @@ gulp.task('watch', ['build'], function() {
 gulp.task('build', ['sass', 'webfont', 'copy:fonts', 'processHtml', 'copy:images']);
 
 // Define Dist generation task (Deploy)
-gulp.task('dist', ['sass:dist', 'fonts:dist', 'processHtml:dist', 'images:dist']);
+gulp.task('dist', ['sass:build', 'fonts:dist', 'processHtml:dist', 'images:dist']);
```
cbernat: You are using sass:build tasks that has sourcemaps build, I think it's best to leave sass:dist here.

gulpfile.js
IgnacioRodrigues: Are you sure this works correctly? I deleted it because it was not working properly for me.

build-w-docker.sh
```diff
@@ -0,0 +1,25 @@
...
+# reference: http://www.ostricher.com/2014/10/the-right-way-to-get-the-directory-of-a-bash-script/
+cd $(dirname $0)
+
+# To avoid issues with MINGW y Git Bash, see:
+# https://github.com/docker/toolbox/issues/673
+# https://gist.github.com/borekb/cb1536a3685ca6fc0ad9a028e6a959e3
+export MSYS_NO_PATHCONV=1
+export MSYS2_ARG_CONV_EXCL="*"
+
+rm -rf dev
```
andresmoschini: It is not necessary because this folder is not mapped inside container (my fault sorry)

build-w-docker.sh
IgnacioRodrigues: It is necessary to start the virtual containers, this activates the updates when changes are made to the files.

andresmoschini:
test this please

gulpfile.js
IgnacioRodrigues: You're right, the change I made does not correspond. I changed it because I was not copying the css  the dev folder to the dist folder. Thanks for the gesture.

build-w-docker.sh
IgnacioRodrigues: Only dev folder is not mapped inside container? Or dev and dist folder? Thanks.

andresmoschini: Only _dist_ folder is mapped:

image

IgnacioRodrigues: You're right, thanks Andres!

gulpfile.js
IgnacioRodrigues: It's okey, we already reviewed it with @cbernat

build-w-docker.sh
andresmoschini: Both of you are fine. But, with a minor detail: it is necessary **for our watch** in the virtual container.

It is because _watch_ need to detect when you made changes, and with this variable, we are forcing use of a polling mechanism.

So, in this case, since we are not watching, please remove it.

andresmoschini:
test this please

andresmoschini:
test this please

PR #29, opened 2019-02-21
fix(build): fixed generation the fonts in gulpfile

IgnacioRodrigues:


cbernat:
Seems to be fine. Can you give a little more detail regarding what the issue was about?

build-w-docker.sh
```diff
@@ -13,6 +13,7 @@ cd $(dirname $0)
 export MSYS_NO_PATHCONV=1
 export MSYS2_ARG_CONV_EXCL="*"
 
+rm -rf dev
```
andresmoschini: I do not understand how could it affect. Could you give us some details?

gulpfile.js
```diff
@@ -82,9 +82,9 @@ var config = {
   }
 };
 
-// Sass tasks are divided for performance issues regarding dependencies
-// Sass Build task definition, only ran once
-gulp.task('sass:build', ['webfont'], function() {
+//Sass tasks are divided for performance issues regarding dependencies
+// Sass Watch task definition
+gulp.task('sass', function() {
```
andresmoschini: If you restore this task to his original position (after `sass:dist`), it would be easier to review the changes. (let me know if I cannot explain myself)

build-w-docker.sh
IgnacioRodrigues: This is because we compare the compilation generated in the dev folder, and it is necessary to eliminate it.

gulpfile.js
IgnacioRodrigues: This review with Andres, and restore the task to original position.

IgnacioRodrigues:
> Seems to be fine. Can you give a little more detail regarding what the issue was about?

In this pull request, what was done was to eliminate a sass task that was not being used, and to test that the flow worked correctly when running the dist task.

PR #33, opened 2019-02-25
fix margins header

GusBaamonde:
fix: margins and paddings header
feat: add styles for scrollbar navigation

### DIFF
--- a/assets/scss/core/_reset.scss
+++ b/assets/scss/core/_reset.scss
@@ -25,6 +25,22 @@ RESET DOPPLER STYLE GUIDES
   box-sizing: border-box;
 }
 
+::-webkit-scrollbar-thumb {
+  background: rgba(0, 0, 0, 0.3);
+  border-radius: 10px;
+}
+
+::-webkit-scrollbar-track {
+  background: rgba(0, 0, 0, 0.1);
+  border-radius: 10px;
+  margin: 5px;
+}
+
+::-webkit-scrollbar {
+  width: 5px;
+  height: 5px;
+}
+
 html {
   line-height: 1.15;
   /* 1 */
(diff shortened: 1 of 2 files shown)


assets/scss/core/_reset.scss
```diff
@@ -25,6 +25,22 @@ RESET DOPPLER STYLE GUIDES
...
+  border-radius: 10px;
+}
+
+::-webkit-scrollbar-track {
+  background: rgba(0, 0, 0, 0.1);
+  border-radius: 10px;
+  margin: 5px;
+}
+
+::-webkit-scrollbar {
```
andresmoschini: Does this changes only affect webkit? is that fine?

andresmoschini:
Thanks @GusBaamonde 

If I am not wrong there are still pending details, but in my opinion it is enough good for the moment.

image

assets/scss/core/_reset.scss
```diff
@@ -25,6 +25,22 @@ RESET DOPPLER STYLE GUIDES
   box-sizing: border-box;
 }
 
+::-webkit-scrollbar-thumb {
+  background: rgba(0, 0, 0, 0.3);
```
cbernat: This color maybe should be set as a variable right?

assets/scss/core/_reset.scss
cbernat: I agree with this, why are you styling webkit scrollbar only? What about the other browsers? Maybe I'm missing something.

andresmoschini: It seems that this _Doppler's issue_ only affects webkit, so, it is fine fixing webkit.

assets/scss/core/_reset.scss
GusBaamonde: Esto hasta quizás se modifique y no lo vamos a usar en ningún otro lado. Por el momento lo dejaría asi.

PR #35, opened 2019-02-25
feat: create separate templates for each component

GusBaamonde:
templates for each component

### DIFF
--- a/assets/templates/messages-header.html
+++ b/assets/templates/messages-header.html
@@ -1,0 +1,57 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <title>Doppler UI Team</title>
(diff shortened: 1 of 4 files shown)


andresmoschini:
What do you think about adding links in the index page? 

Right now I am only seeing some colors:

image

assets/templates/messages-header.html
```diff
@@ -0,0 +1,57 @@
...
+  </div>
+</div><!-- error -->
+<!-- amarillo -->
+<div class="messages-container warning">
+  <div class="wrapper">
+    <p>Thanks! We’ll contact you ASAP and if everything is fine, you’ll be able to send your Campaign.</p>
+  </div>
+</div><!-- warning-->
+<!-- verde -->
+<div class="messages-container message">
```
fcoronelmakingsense: can add css class 'info' with the same styles that 'message'. We need for the webapp. Thanks!

andresmoschini:
image

Please, try to rebase in place of backward merge.

PR #39, opened 2019-02-27
feat: add styles for modal and form user-menu

GusBaamonde:
Standard modal sizes creation and form for user-menu


assets/js/main.js
```diff
@@ -3,21 +3,36 @@ $("body").on({
...
+$("body").on('click', '#close-menu', function() {
   $('.header-main').removeClass('open');
-});
\ No newline at end of file
+});
+
+// modals
+
+$("#btn-modal-large").click(function() {
+  $("#modal-large").fadeIn(500);
```
cbernat: Maybe add this default delay into a var `500` What do you think?

GusBaamonde: great idea!

GusBaamonde:
> Good work Gus



> Good work Gus

Thanks Nacho!

PR #44, opened 2019-03-01
fix: modal interaction

fcoronelmakingsense:
- Initialize all modals hidden
- Add close user menu when click in upgrade plan button

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -23,8 +23,15 @@ $("body").on('click', '#close-menu', function() {
   $('.header-main').removeClass('open');
 });
 
+$("body").on('click', '.close-user--menu', function() {
+  $('.user-menu').removeClass('open');
+});
+
 // modals
 var $delayFade = 500;
+//Initialize all modals hidden
+$("#modal-large, #modal-medium, #modal-small").hide();
+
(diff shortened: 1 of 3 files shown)


assets/js/main.js
```diff
@@ -23,8 +23,15 @@ $("body").on('click', '#close-menu', function() {
   $('.header-main').removeClass('open');
 });
 
+$("body").on('click', '.close-user--menu', function() {
+  $('.user-menu').removeClass('open');
```
andresmoschini: It closes the menu when the user clicks on, for example, "COMPRAR". But should you also close the menu when the user clicks outside the menu?

image

assets/js/main.js
```diff
@@ -23,8 +23,15 @@ $("body").on('click', '#close-menu', function() {
...
+});
+
 // modals
 var $delayFade = 500;
+//Initialize all modals hidden
+$("#modal-large, #modal-medium, #modal-small").hide();
```
andresmoschini: 👍

assets/js/main.js
fcoronelmakingsense: This is another thing we need to add. I add this for the current modal we need

cbernat:
👍

PR #47, opened 2019-03-06
fix: header nav right to avoid inherit ul styles

fcoronelmakingsense:
- fix header nav right to avoid inherit ul styles in modal on webapp

### DIFF
--- a/assets/scss/templates/_header.scss
+++ b/assets/scss/templates/_header.scss
@@ -121,14 +121,11 @@
   display: flex;
 }
 
-.nav-right-main > ul {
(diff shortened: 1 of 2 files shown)


andresmoschini:
I think that it is fine, but it would be nice if @GusBaamonde, @damianmuti, @IgnacioRodrigues, @lucianaDilizio, @mvillaseco or @cbernat are better to review this kind of changes.

cbernat:
👍

IgnacioRodrigues:
It's okey Fede! 👍

PR #54, opened 2019-03-15
Adding disabled and grey button styles

GusBaamonde:
Adding disabled and grey button styles

See demo in https://cdn.fromdoppler.com/doppler-ui-library/build97/

image

Updated: 

image

### DIFF
--- a/assets/scss/modules/_buttons.scss
+++ b/assets/scss/modules/_buttons.scss
@@ -50,6 +50,17 @@
           border: 1px solid $dp-color-darkgreen;
         }
       }
+
+      &grey {
+        background-color: $dp-color-grey;
+        border: 1px solid $dp-color-grey;
+        color: $dp-color-white;
+
+        &:hover {
+          background-color: $dp-color-darkgrey;
+          border: 1px solid $dp-color-darkgrey;
+        }
+      }
     }
   }
 
--- a/assets/scss/modules/_buttons.scss
+++ b/assets/scss/modules/_buttons.scss
@@ -52,5 +52,13 @@
       }
     }
   }
+
+  .dp-button:disabled {
+    pointer-events: none;
+    border: 1px solid $dp-color-sylver;
+    color: $dp-color-grey;
+    background-color: $dp-color-snow;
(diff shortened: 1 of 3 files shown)


GusBaamonde:
- Add styles for button disabled
- Add class primary-grey for button

[](https://cdn.fromdoppler.com/doppler-ui-library/build97/index.html)

assets/scss/modules/_buttons.scss
```diff
@@ -50,7 +50,26 @@
...
+
+        &:hover {
+          background-color: $dp-color-darkgrey;
+          border: 1px solid $dp-color-darkgrey;
+        }
+      }
     }
   }
+
+  .dp-button:disabled {
```
andresmoschini: Only to learn, how do you guarantee that these styles have priority over color styles?

andresmoschini:
Nice work! remember to add a new PR at webapp:

* Updating the references in https://github.com/FromDoppler/doppler-webapp/blob/master/public/index.html#L14
* And using the right styles in https://github.com/FromDoppler/doppler-webapp/blob/master/src/components/UpgradePlanForm/UpgradePlanForm.js#L85-L95

assets/scss/modules/_buttons.scss
GusBaamonde: in this case, it works because of the cascade effect and not for the CSS specificity due to the class color and the disabled class color have the same weight (specificity)

andresmoschini:
hi @GusBaamonde, reading [a react forms helper library documentation](https://jaredpalmer.com/formik/), I have noticed that in many examples they seem to apply _disabled_ state on submitting.

So, if we apply the same criteria, we will have buttons with _disabled_ and _loading_ status. Did you verify how both styles work together?

Let me know if I cannot explain myself.

GusBaamonde:
They work perfectly, I already tried it. Maybe we have to change the spinner color to make it more visible.
Now I understand why mailchimp uses it that way, they also use react.

cbernat:
👍

andresmoschini:
Updated: 
image

PR #55, opened 2019-03-18
DBR-91 Fix issues with user menu

damianmuti:
fix: add a way to close user menu when clicking outside it

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -27,6 +27,14 @@ $("body").on('click', '.close-user--menu', function() {
   $('.user-menu').removeClass('open');
 });
 
+$(document).on('mouseup', function(e){
+	var menu = $(".user-menu");
+
+	if (!menu.is(e.target) && menu.has(e.target).length === 0){
+		menu.removeClass('open');
+	}
+}); 
+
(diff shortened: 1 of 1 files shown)


assets/js/main.js
```diff
@@ -27,6 +27,14 @@ $("body").on('click', '.close-user--menu', function() {
   $('.user-menu').removeClass('open');
 });
 
+$(document).on('mouseup', function(e){
+	var menu = $(".user-menu");
+
+	if (!menu.is(e.target) && menu.has(e.target).length === 0){
+		menu.removeClass('open');
+	}
```
andresmoschini: If I am not wrong we were using spaces in place of tabs.

andresmoschini:
Thanks! I am not so much familiar with DOM manipulation, but it looks to be fine.

assets/js/main.js
damianmuti: I have the code editor set up in 2 spaces, but this would be a great opportunity to create a `.editorconfig` file to normalize how we all create code for the project. 👍

andresmoschini:
🚀

PR #57, opened 2019-03-19
Refactor components in the form signup

IgnacioRodrigues:
fix(signup): refactor components in the form signup

CDN:
http://cdn.fromdoppler.com/doppler-ui-library/build105/signup.html

### DIFF
--- a/assets/scss/modules/_forms.scss
+++ b/assets/scss/modules/_forms.scss
@@ -1,4 +1,5 @@
 /* form-request - user-menu */
+
 .form-request {
   font-family: $dp-font-family;
   padding: 20px 0 0;
@@ -16,14 +17,14 @@
   }
 }
 
-.content-checkbox {
+.checkbox-group {
   width: 100%;
   position: relative;
   display: inline-block;
   text-align: left;
   margin: 20px 0;
 
-  .checkbox {
+  .holder-checkbox {
     display: flex;
     cursor: pointer;
     align-items: center;
@@ -55,6 +56,35 @@
       border: solid $dp-color-white;
       border-width: 0 3px 3px 0;
       transform: rotate(45deg);
+    }
+  }
+}
+
+.holder {
+  &-group {
+    display: flex;
+    flex-wrap: wrap;
+    flex-direction: column;
+  }
+
+  &-size1of1 {
+    flex: 1;
+
+    &:nth-last-child(-n+3) {
+      margin-top: 30px;
+    }
+  }
+
+  &-col {
+    flex-wrap: nowrap;
+    flex-direction: row;
+  }
+
+  &-size1of2 {
+    flex: 1;
+
+    &:nth-child(1) {
+      margin-right: 4%;
     }
   }
 }
--- a/assets/scss/templates/_signup.scss
+++ b/assets/scss/templates/_signup.scss
@@ -40,6 +40,7 @@
     .content-subtitle {
       margin: 10px 0 30px;
       font-family: $dp-font-family-proximanova;
+      font-style: italic;
     }
 
     .content-promotion {
@@ -48,18 +49,36 @@
       font-family: $dp-font-family-proximanova;
     }
 
-    .signup-form {
-      .holder-small {
-        display: inline-block;
-        width: 47.4%;
+    .content-legal {
+      margin-top: 30px;
+    }
+
+    .holder {
+      &-group {
+        display: flex;
+        flex-wrap: wrap;
+        flex-direction: column;
+      }
+
+      &-size1of1 {
+        flex: 1;
+
+        &:nth-last-child(-n+3) {
+          margin-top: 30px;
+        }
+      }
+
+      &-col {
+        flex-wrap: nowrap;
+        flex-direction: row;
+      }
+
+      &-size1of2 {
+        flex: 1;
 
         &:nth-child(1) {
           margin-right: 4%;
         }
-      }
-
-      ul > li:nth-child(n+3) {
-        margin-top: 20px;
       }
     }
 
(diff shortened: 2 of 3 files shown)


assets/scss/modules/_forms.scss
```diff
@@ -16,14 +17,14 @@
...
-.content-checkbox {
+.checkbox-group {
   width: 100%;
   position: relative;
   display: inline-block;
   text-align: left;
   margin: 20px 0;
 
-  .checkbox {
+  .holder-checkbox {
```
cbernat: I see there is a difference in the way you made this clases `.checkbox-group` and `.holder-checkbox` why is that? Perhaps the second class may be `.checkbox-holder` What do you think?

assets/scss/modules/_forms.scss
```diff
@@ -57,4 +58,33 @@
...
+  &-group {
+    display: flex;
+    flex-wrap: wrap;
+    flex-direction: column;
+  }
+
+  &-size1of1 {
+    flex: 1;
+
+    &:nth-last-child(-n+3) {
```
cbernat: Is this ok? ` (-n+3)`

assets/scss/modules/_forms.scss
```diff
@@ -57,4 +58,33 @@
...
+  &-col {
+    flex-wrap: nowrap;
+    flex-direction: row;
+  }
+
+  &-size1of2 {
+    flex: 1;
+
+    &:nth-child(1) {
+      margin-right: 4%;
```
GusBaamonde: Strange measure warning!

assets/scss/templates/_signup.scss
```diff
@@ -48,18 +49,36 @@
...
+      &-col {
+        flex-wrap: nowrap;
+        flex-direction: row;
+      }
+
+      &-size1of2 {
+        flex: 1;
+
+        &:nth-child(1) {
+          margin-right: 4%;
```
GusBaamonde: Strange measure warning!

assets/scss/modules/_forms.scss
IgnacioRodrigues: Yes, is an option @cbernat  may be ´.checkbox-holder´ is the nomenclature correct! Thanks 👍

assets/scss/modules/_forms.scss
IgnacioRodrigues: Yes, it's okay. Can you think of something better?

assets/scss/modules/_forms.scss
IgnacioRodrigues: What is the problem?

assets/scss/templates/_signup.scss
IgnacioRodrigues: what is the problem?

assets/scss/modules/_forms.scss
IgnacioRodrigues: May be use ´justify-content: space-between;´

assets/scss/templates/_signup.scss
damianmuti: @IgnacioRodrigues seems to be a magic number. It would be much better if we use an absolute unit here 👌

assets/templates/signup.html
```diff
@@ -37,42 +37,46 @@
...
             </li>
-            <li>
+            <li class="holder-size1of1">
               <label for="email">Email:</label>
               <input type="email" name="email" placeholder="Tu Email será tu nombre de usuario">
             </li>
-            <li>
+            <li class="holder-size1of1">
               <label for="password">Contraseña:</label>
               <input type="password" name="password" placeholder="Escribe tu clave secreta">
```
damianmuti: Inputs should have the `id` attribute with a value that matches the `name` attribute of it's label for the label to be clickeable (so it can focus on the input). Same goes with all of the inputs.

assets/templates/signup.html
```diff
@@ -37,42 +37,46 @@
...
+            <li class="holder-checkbox">
               <input type="checkbox" name="acepto_politicas">
               <span class="checkmark"></span>
               <label for="acepto_politicas">Acepto la <a class="link-green" href="#">Política de Privacidad</a> de Doppler.</label>
             </li>
-            <li class="checkbox">
+            <li class="holder-checkbox">
               <input type="checkbox" name="acepto_promociones">
               <span class="checkmark"></span>
               <label for="acepto_promociones">Acepto las promociones de Doppler y sus aliados.</label>
```
damianmuti: Label baseline should be the bottom of the checkbox, so it can be aligned properly.

image

assets/scss/templates/_signup.scss
```diff
@@ -48,18 +49,36 @@
...
+      &-group {
+        display: flex;
+        flex-wrap: wrap;
+        flex-direction: column;
+      }
+
+      &-size1of1 {
+        flex: 1;
+
+        &:nth-last-child(-n+3) {
```
damianmuti: Why target the last 3 rows specifically?

assets/templates/signup.html
```diff
@@ -37,42 +37,46 @@
...
-            <li>
+            <li class="holder-size1of1">
               <label for="password">Contraseña:</label>
               <input type="password" name="password" placeholder="Escribe tu clave secreta">
             </li>
           </ul>
-        </fieldset>
-        <fieldset class="holder-policy">
+        </fieldset><!-- personal information, user and password -->
+        <fieldset>
```
damianmuti: Is there a specific reason to have 2 different fieldsets in the same form?

assets/scss/templates/_signup.scss
IgnacioRodrigues: Ok!

assets/templates/signup.html
IgnacioRodrigues: Ok!

assets/templates/signup.html
IgnacioRodrigues: It is not possible, because the input was replaced by a span that contains the correct design. What you can do is put a height so that it does not align itself.

assets/scss/templates/_signup.scss
IgnacioRodrigues: Yes, I know what you mean. It's maybe replace for &:nth-child(n+2)

assets/templates/signup.html
IgnacioRodrigues: There is no special reason why we add in the 2 different fieldsets in the same form, this can be separated if we want

IgnacioRodrigues:
I'm going to close this pull request because there are many things to correct.

PR #58, opened 2019-03-19
Fix Signup

IgnacioRodrigues:
Fix styles in component the signup and styles 

CDN: 
http://cdn.fromdoppler.com/doppler-ui-library/build123/signup.html


assets/scss/core/_typography.scss
```diff
@@ -25,7 +25,9 @@ $dp-headers-h4-font-size: 22px;
 $dp-form-font-size: 20px;
 $dp-headers-h5-font-size: 18px;
 $dp-headers-h6-font-size: 13px;
-$dp-font-size-small: 13px;
+$dp-font-size-small: 14px;
```
andresmoschini: Does this change not affect other pages?

assets/scss/styles.scss
```diff
@@ -21,6 +21,7 @@
 @import "core/variables";
 @import "core/base";
 @import "libs/**/*";
-@import "modules/**/*";
+@import "modules/mixins/*";
+@import "modules/*";
```
andresmoschini: Is it fine?

assets/templates/signup.html
```diff
@@ -40,54 +40,62 @@ <h5>Regístrate</h5>
...
+        <fieldset>
+          <ul class="holder-group">
+            <legend>Usuario y contraseña</legend>
+            <li class="holder-item holder-item--big">
               <label for="email">Email:</label>
-              <input type="email" name="email" placeholder="Tu Email será tu nombre de usuario">
+              <input type="email" name="email" id="email" placeholder="Tu Email será tu nombre de usuario">
             </li>
-            <li>
+            <li class="holder-item holder-item--big">
```
andresmoschini: @GusBaamonde should we use these classes in place of `field-group` in #59 ?

andresmoschini:
@IgnacioRodrigues could you add this signup page to the navigation bar?

assets/scss/core/_typography.scss
IgnacioRodrigues: Yes, it's talking with Santiago

assets/scss/styles.scss
IgnacioRodrigues: Yes! We review this with Gus

IgnacioRodrigues:
> @IgnacioRodrigues could you add this signup page to the navigation bar?

You're right! Tomorrow I add it to the navigation header.

assets/scss/core/_typography.scss
mvillaseco: This are the current sizes:
11px(x2),12px,13px(x2),14px(x2),15px(x2),16px,18px(x2),20px,22px(x),24px,30px,40px,50px.

Clearly we have a problem here.

assets/scss/modules/_forms.scss
```diff
@@ -16,45 +17,110 @@
...
+  float: left;
+  box-sizing: border-box;
   text-align: left;
-  margin: 20px 0;
 
-  .checkbox {
-    display: flex;
+  &--small {
+    width: 50%;
+    margin-bottom: 30px;
```
mvillaseco: Great place to start using lvls (a variable that we can change after when santi have the real values)

assets/scss/modules/_forms.scss
```diff
@@ -16,45 +17,110 @@
...
+    // Left Column
+    &:nth-of-type(2n+1) {
+      padding-right: 20px;
+    }
+  }
+
+  &--big {
+    width: 100%;
+    margin-bottom: 30px;
```
mvillaseco: same lvl here

assets/scss/modules/_forms.scss
```diff
@@ -16,45 +17,110 @@
...
-  margin: 20px 0;
 
-  .checkbox {
-    display: flex;
+  &--small {
+    width: 50%;
+    margin-bottom: 30px;
+    // Left Column
+    &:nth-of-type(2n+1) {
+      padding-right: 20px;
```
mvillaseco: lower lvl here (maybe 10px, 20px, 30px and 40px);

assets/scss/modules/_forms.scss
```diff
@@ -16,45 +17,110 @@
...
   position: relative;
-  display: inline-block;
+  float: left;
+  box-sizing: border-box;
   text-align: left;
-  margin: 20px 0;
 
-  .checkbox {
-    display: flex;
+  &--small {
```
mvillaseco: not sure if this is ok

assets/scss/modules/_forms.scss
```diff
@@ -16,45 +17,110 @@
...
+
+    input,
+    textarea {
+      width: 100%;
+    }
+  }
+
+  label {
+    color: $dp-color-grey;
+    margin-bottom: 0;
```
mvillaseco: i'm pretty sure, gustavo added a margin 0 variable or class

assets/scss/modules/_forms.scss
```diff
@@ -16,45 +17,110 @@
...
+    vertical-align: baseline;
+  }
+
+  &.checkbox {
     cursor: pointer;
+    display: flex;
     align-items: center;
+    margin-bottom: 0;
```
mvillaseco: same as before :)

assets/scss/modules/_forms.scss
```diff
@@ -16,45 +17,110 @@
...
+
+    &:nth-last-child(-n+1) {
+      margin-bottom: 20px;
```
mvillaseco: variable?

assets/scss/core/_typography.scss
IgnacioRodrigues: Yes, missing better definition of font sizes

assets/scss/modules/_forms.scss
IgnacioRodrigues: Yes! It's true, maybe in this case add the class for de margin

assets/scss/templates/_signup.scss
```diff
@@ -6,7 +6,7 @@
   align-items: stretch;
 
   .main-panel {
-    flex: 1.5;
+    width: 500px;
```
mvillaseco: hmmm, this is not going to increase or decrease it size.... is this ok?

assets/templates/signup.html
```diff
@@ -40,54 +40,62 @@ <h5>Regístrate</h5>
...
+                <li class="holder-item holder-item--small">
+                  <label for="name">Nombre:</label>
+                  <input type="name" name="name" id="name" placeholder="Nombre">
+                <li class="holder-item holder-item--small">
+                  <label for="lastname">Apellido:</label>
+                  <input type="lastname" name="name" id="lastname" placeholder="Apellido">
+                </li>
+              </ul>
+            </li>
+            <li class="holder-item holder-item--big">
```
mvillaseco: i'm not sure but, how about having a "holder-item" main class and have "big" and "small" modifiers classes to only adjust the size?

assets/scss/templates/_signup.scss
IgnacioRodrigues: Yes! This section no going to increase or decrease the size, maintain the same width and the breakpoint change to mobile

IgnacioRodrigues:
The styles of the components were added in the corresponding files, classes were added for the spaces (which @GusBaamonde  will add to variables.scss) and styles that were not good were eliminated.

assets/templates/signup.html
```diff
@@ -41,46 +41,46 @@ <h5>Regístrate</h5>
...
               <input type="password" name="password" id="password" placeholder="Escribe tu clave secreta">
             </li>
           </ul>
         </fieldset>
         <fieldset>
           <legend>Politica de privacidad y promociones</legend>
-          <ul class="holder-group">
-            <li class="holder-item holder-item--big checkbox">
+          <ul class="field-group">
+            <li class="field-item field-item__checkbox m-bottom--lv5"><!-- margin-bottom: 30px m-bottom--lv5 -->
```
cbernat: What's the difference between this `field-item__checkbox` and this `m-bottom--lv5` in terms of use of BEM, maybe we should stick to only one way of use. What do you think?

IgnacioRodrigues: The difference they have is that "--lv-5" is a modifier and "__checkbox" is an element the field-item, the BEM convention says that it is block__element--modifier, with the name the block or element or modifier separate with a "-"

assets/scss/core/_base.scss
```diff
@@ -78,15 +76,120 @@ textarea {
...
+        vertical-align: middle;
+        margin-left: 25px;
+      }
+    }/* Field checkbox */
+
+    &--50 {
+      width: 50%;
+      /* Left Column */
+      &:nth-of-type(2n+1) {
+        padding-right: 20px;
```
andresmoschini: Are you still using arbitrary spacings? or it is missing take advantage of Gustavo variables?

assets/scss/core/_variables.scss
```diff
@@ -85,6 +86,9 @@ $dp-color-brown-brightness: rgba(48,33,0,.2); //
 $dp-border-radius: 3px;
 $dp-header-height: 4em;
 $dp-text-aling: center;
+//
```
andresmoschini: What does this comment mean?

assets/scss/templates/_signup.scss
```diff
@@ -124,28 +119,21 @@
...
-        }
+      width: 100%;
 
-        ul > li:nth-child(n+2) {
-          margin-top: 20px;
+      .field-item--50 {
+        width: 100%;
+        // Left Column
+        &:nth-of-type(2n+1) {
+          padding-right: 0;
```
andresmoschini: fix still pending, right?

assets/scss/core/_base.scss
IgnacioRodrigues: This was already corrected in the next commit 👍

assets/scss/templates/_signup.scss
IgnacioRodrigues: Yes, this was already corrected in the next commit

IgnacioRodrigues:
CDN:
http://cdn.fromdoppler.com/doppler-ui-library/build128/signup.html

A mixin was added for the validation circles in the field of the password indicated by the color. Fixed errors such as magic numbers and the use of variables were corrected.

assets/scss/templates/_signup.scss
andresmoschini: Perdón, no entendí, _fue_ corregido o _va a ser_ corregido? en el último caso, dentro de este mismo pr o en otro?

IgnacioRodrigues: Si, ya lo corregí en este pull request, mas a bajo vas a ver el commit

andresmoschini: En ese caso, la variable `$dp-spaces--lv0` no tiene sentido, deberíamos borrarla?

https://github.com/FromDoppler/Doppler-UI-System/pull/59/files#diff-c6347520a93a62f1f778dd66f6bba5d3R42

@GusBaamonde @mvillaseco

GusBaamonde: Seguramente la necesitaremos, sobre todo si por herencia los elementos generan margins o paddings, esta variable nos ayudaría de mucho.

IgnacioRodrigues: Ok, entonces la agrego ;)

PR #59, opened 2019-03-19
Validate textarea

GusBaamonde:
- add animation to show the errors (optional)
- add classes to validate form
- add elements and class html
- add class for modal button layout

form_validation


andresmoschini:
Looks nice!

@FromDoppler/doppler-webapp please review naming conventions! I am not sure if it is compatible with @IgnacioRodrigues's PR #58 

@GusBaamonde could you explain to me how to apply the animation?

assets/scss/core/_base.scss
```diff
@@ -68,10 +66,88 @@ textarea {
...
+    border: 1px solid $dp-color-red;
+  }
+
+  .error-messages {
+    display: inline-block;
+
+    p {
+      margin: 0;
+      color: $dp-color-red;
+      font-size: $dp-font-size-small;
```
andresmoschini: What do you think about adding these properties to `.error-messages` and not only to `error-messages p`?

I do not know if `p` will be always the right way of representing this information.

fcoronelmakingsense:
Maybe is better validate with the Designer the animation because in doppler doesn't has animation on validations.

assets/scss/core/_base.scss
```diff
@@ -31,9 +29,9 @@ label {
   border-color: $dp-color-lightgrey transparent transparent transparent;
   position: absolute;
   right: 15px;
-  top: 67%;
+  top: 2.5em;
```
fcoronelmakingsense: define in ems depends of the parent value right? it's not a problem this? in any case?

assets/scss/templates/_guidelines.scss
```diff
@@ -132,6 +132,15 @@
     color: $dp-color-darkgrey;
   }
 }
+/* modals */
+.list-modal {
+  display: flex;
+  justify-content: space-around;
+  min-height: 200px;
+  align-items: center;
+  max-width: 1200px;
```
fcoronelmakingsense: this value it's not defined in global file?

assets/scss/core/_base.scss
mvillaseco: We are talking about an arrow inside a dropdown, it should depend on the parent always, at least that's my opinion, of course i can be wrong about this.

assets/scss/core/_base.scss
```diff
@@ -68,10 +66,88 @@ textarea {
...
+    border: 1px solid $dp-color-yellow;
+  }
+}
+/* validate form*/
+.field-group {
+  position: relative;
+  display: inline-block;
+  text-align: left;
+  width: 100%;
+  margin-bottom: 30px;
```
mvillaseco: Nacho told me that you are going to add the space lvls in another PR right?, maybe this line should be removed

andresmoschini: As we have already talked with @mvillaseco, he will add in this same PR.

Basically, constants to be used here, and atomic classes for other cases.

assets/scss/modules/_atomic-class.scss
```diff
@@ -0,0 +1,54 @@
...
+$dp-m-bottom--lv6: 36px;
+$dp-m-bottom--lv7: 42px;
+$dp-m-bottom--lv8: 48px;
+/* atomic class for margin */
+.margin--lv0 {
+  margin: 0;
+}
+
+.margin--lv1 {
+  margin: 6px;
```
andresmoschini: Is it not possible using the variable here?

assets/scss/modules/_atomic-class.scss
```diff
@@ -0,0 +1,54 @@
...
+}
+/* atomic class for margin bottom */
+.m-bottom--lv0 {
+  margin-bottom: 0;
```
andresmoschini: Is it not possible using the variable here (an in the rest of the cases)?

assets/scss/modules/_atomic-class.scss
andresmoschini: Is it fine having this one? it does not have a related variable.

GusBaamonde:
- Por semántica decidimos utilizar “field-group” como padre y cada li  como hijo “field-item”.
- Creamos un archivo con clases atómicas y variables para la normalización de espacios. (como decían @mvillaseco @damianmuti ) 
- Para las validaciones arme un contenedor (wrapper-errors) y dentro la clase (error-message)
- Coloqué dos elementos diferentes (p y span) para ver su comportamiento.
- La animación bounceIn la dejaremos en stand by para futuras discuciones. Como bien dijo @fcoronelmakingsense  para ser consistentes con el resto de la aplicación.

form01

assets/scss/modules/_atomic-class.scss
```diff
@@ -0,0 +1,54 @@
+/* variables margin bottom */
+$dp-m-bottom--lv0: 0;
+$dp-m-bottom--lv1: 6px;
```
andresmoschini: What do you think about naming it  `$dp-m--lv1`, I mean, all margins _lv1_ will have 6px, right? or does it care if it is a top, botton, left or right marging?

assets/scss/styles.scss
```diff
@@ -14,11 +14,11 @@
 ///
 /// -> Utilities
 /// -
+@import "core/variables";
 @import "core/colors";
 @import "core/reset";
 @import "core/settings";
 @import "core/typography";
-@import "core/variables";
```
andresmoschini: Is this change on purpose?

GusBaamonde: change the order, you need to load the variables first to be able to use them

assets/scss/core/_base.scss
andresmoschini: @GusBaamonde this change is still pending, right?

Feel free to call @mvillaseco if it is not clear.

assets/scss/modules/_atomic-class.scss
GusBaamonde: No, los margenes son top, right, bottom, left y en algunos casos el margen es para los cuatro valores. por eso $dp-m-bottom es diferente a $margin--lv0 (que me olvide de agregar)

andresmoschini: Me parece que no. Mejor lo chateamos.

assets/scss/styles.scss
damianmuti: If we follow our framework pattern, variables should be included in the settings file.

assets/scss/modules/_atomic-class.scss
damianmuti: Wouldn't it be better to create a mixin that can handle margin/padding positioning and value instead of creating each of the classes manually?

assets/scss/styles.scss
GusBaamonde: > If we follow our framework pattern, variables should be included in the settings file.

Si, tenemos que refactorizar las variables

assets/scss/modules/_atomic-class.scss
GusBaamonde: I do not know how to do that @damianmuti , do you have any examples at hand?

andresmoschini: Regarding the mixings, in this case, I think that right now it is enough simple. So, if you have time to talk about it and use the mixings, it is fine. If not, keep this improvement for the future.

Maybe @damianmuti could generate a little PR with the refactoring after your merge.

assets/scss/styles.scss
andresmoschini: Creo que @damianmuti simplemente propone mover la definición de las variables a otro archivo, creo que es un cambio simple.

Si no está claro, lo podemos dejar así y luego @damianmuti puede hacer un PR rápido para corregirlo.

andresmoschini:
Something is breaking here: https://cdn.fromdoppler.com/doppler-ui-library/build113/

andresmoschini:
Great! it is working now! 

:shipit: 

Please merge it now. If @damianmuti have time, tomorrow he could apply the suggested refactorings in order to allow us to review them.

PR #60, opened 2019-03-20
Dbr 100 user menu bug

damianmuti:
Besides removing the styles that were unused and were affecting the structure for the user menu, I've created a mixin to handle ellipsis on elements.
This mixin is intended for internal usage, meaning that this will help Library creators do ellipsis on elements based on given attrs.

### DIFF
--- a/assets/scss/templates/_header.scss
+++ b/assets/scss/templates/_header.scss
@@ -239,14 +239,6 @@
     align-items: center;
   }
 
-  span {
-    text-overflow: ellipsis;
-    white-space: nowrap;
-    max-width: 248px;
-    display: block;
-    overflow: hidden;
-  }
-
   header p {
     margin-left: 10px;
   }
--- a/assets/scss/templates/_header.scss
+++ b/assets/scss/templates/_header.scss
@@ -195,10 +195,7 @@
   }
 }
 
-.user-menu.open {
-  visibility: visible;
-  opacity: 1;
-}
+
 
 .user-menu {
   background: $dp-color-white;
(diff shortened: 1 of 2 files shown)


assets/scss/templates/_header.scss
```diff
@@ -239,14 +239,6 @@
...
   }
 
-  span {
-    text-overflow: ellipsis;
-    white-space: nowrap;
-    max-width: 248px;
-    display: block;
-    overflow: hidden;
-  }
-
```
andresmoschini: Is this code really redundant? @damianmuti please, verify all demo pages in the library and update reference in webapp before merge it.

andresmoschini: image

This issue seems to be related to your change, could you verify?

damianmuti: Added ellipsis back on the specific elements that needed it. Good catch!

assets/scss/templates/_header.scss
```diff
@@ -234,45 +231,46 @@
...
-    overflow: hidden;
-  }
+    p {
+      margin-left: 10px;
 
-  header p {
-    margin-left: 10px;
-  }
+      span {
+        @include do-ellipsis(248px, block);
```
andresmoschini: Are you taking advantage of the mixing here, right? ❤️

damianmuti: Indeed, @andresmoschini 🙌

assets/scss/utilities/_ellipsis.scss
```diff
@@ -0,0 +1,29 @@
+////
+/// Ellipsis
+/// @group utilities
+/// @access private
+/// @author Damián Muti
+////
+
+/// Create ellipsis on an element.
```
cbernat: Is this the form of documentation you were suggesting @damianmuti ?

assets/scss/templates/_header.scss
```diff
@@ -234,45 +231,46 @@
...
+      color: $dp-color-grey;
+      font-size: $dp-font-size-small;
+    }
 
-  .name {
-    color: $dp-color-darkgrey;
-    font-size: 20px;
+    .name {
+      color: $dp-color-darkgrey;
+      font-size: 20px;
```
cbernat: I think we were using variables for font-size would you mind checking this? Same for `font-size: 11px;` in  `.option-user li`

assets/scss/utilities/_ellipsis.scss
damianmuti: Exactly, @cbernat . It makes use of SassDoc comment structure to parse it and create a separated technical documentation based on it.

assets/scss/templates/_header.scss
damianmuti: @cbernat I did not refactor any of the styles. Just cleaned up indentation and fixed the bug. I guess we should do some refactoring once everything is defined.

andresmoschini:
🚀

PR #64, opened 2019-03-25
SignUp fixs

IgnacioRodrigues:
In this PR, some bugs were modified, such as:
- Focus on checkbox
- The color of the input text
- Doppler logo by class
- Spaces
The eye was added to see the password.

CDN PREVIEW:
http://cdn.fromdoppler.com/doppler-ui-library/build137/signup.html


andresmoschini:
A lot of fixes all together!

The next time, please, try to do small fixes one by one.

I will review as soon as possible. Thanks.

assets/templates/signup.html
```diff
@@ -33,7 +33,7 @@
   <main class="panel-wrapper">
     <article class="main-panel">
       <header>
-        <img src="img/doppler-logo.svg" class="logo-doppler" alt="Doppler">
+        <h1 class="logo-doppler-new">Doppler</h1>
```
andresmoschini: ❤️

assets/scss/core/_base.scss
```diff
@@ -200,7 +237,8 @@ input[type="tel"] {
 .error {
   input,
   select,
-  textarea {
+  textarea,
+  .checkmark {
```
andresmoschini: 👍

assets/scss/core/_base.scss
```diff
@@ -147,12 +172,24 @@ input[type='checkbox'] {
       }
     }
   }
+
+  &:focus {
+    & ~ .checkmark {
```
andresmoschini: ❤️

assets/js/main.js
```diff
@@ -52,3 +52,20 @@ $("#btn-modal-small").click(function() {
 $(".modal .close").click(function() {
   $(".modal").fadeOut();
 });
+
+
+// script to show and hide password
+$(".show-hide").click(function() {
```
andresmoschini: This code will not work in our react webapp, let us talk in order to refactorize it.

assets/js/main.js
```diff
@@ -52,3 +52,20 @@ $("#btn-modal-small").click(function() {
...
+  $(this).toggleClass("ms-icon icon-hide");
+  var input = $($(this).attr("toggle"));
```
andresmoschini: What do you think about making this smarter and search the input with the type password in current _field-item_? (and avoid using the reference)

andresmoschini: Oh, you cannot because after that you convert it into a text, but, well, there are two options:
* in that moment you could store a reference to the input.
* or, directly use the only input in this _field-group_, there should be only one.

assets/js/main.js
```diff
@@ -52,3 +52,20 @@ $("#btn-modal-small").click(function() {
 $(".modal .close").click(function() {
   $(".modal").fadeOut();
 });
+
+
+// script to show and hide password
+$(".show-hide").click(function() {
+  $(this).toggleClass("ms-icon icon-hide");
```
andresmoschini: Some lines later it seems that you update the class again, something should be redundant, right?

assets/js/main.js
```diff
@@ -52,3 +52,20 @@ $("#btn-modal-small").click(function() {
...
+  var input = $($(this).attr("toggle"));
+
+  if (input.attr("type") == "password") {
+    input.attr("type", "text");
+    $('span.content-eye').text('Ocultar');
+    $(this).removeClass("ms-icon icon-view").addClass("ms-icon icon-hide");
+  } else {
+    input.attr("type", "password");
+    $('span.content-eye').text('Mostrar');
```
andresmoschini: could you read this text literals form the DOM in order to allow us to translate it?

assets/js/main.js
andresmoschini: Another option is that we remake this logic using React. In that case, this code is only for the library demo, and should not be here.

IgnacioRodrigues: This code is Jquery is only demonstrative to see how it has to work, it really should be done with React. We could remove it so that it does not interfere.

andresmoschini: > This code is Jquery is only demonstrative to see how it has to work, it really should be done with React. We could remove it so that it does not interfere.

No, it is not. [We are including this code](https://github.com/FromDoppler/doppler-webapp/blob/master/public/index.html#L47) in UI Library clients.

The demonstrative code should be in another place.

IgnacioRodrigues: Ok

assets/js/main.js
IgnacioRodrigues: Yes, it is possible to implement the second item, store a reference to the entry to make the code smarter.

assets/js/main.js
```diff
@@ -52,3 +52,20 @@ $("#btn-modal-small").click(function() {
...
   $(".modal").fadeOut();
 });
+
+
+// This script is ONLY to DEMONSTRATE how to show and hide password, not use in production
+$(".show-hide").click(function() {
+  $(this).toggleClass("ms-icon icon-hide");
+  var input = $($(this).attr("toggle"));
+
+  if (input.attr("type") == "password") {
```
cbernat: We try to use '===' instead of '=='.

assets/js/main.js
```diff
@@ -52,3 +52,20 @@ $("#btn-modal-small").click(function() {
...
+    input.attr("type", "text");
+    $('span.content-eye').text('Ocultar');
+    $(this).removeClass("ms-icon icon-view").addClass("ms-icon icon-hide");
+  } else {
+    input.attr("type", "password");
+    $('span.content-eye').text('Mostrar');
+    $(this).removeClass("ms-icon icon-hide").addClass("ms-icon icon-view");
```
cbernat: I think this `ms-icon` is redundant since you add then again.

assets/scss/modules/_atomic-class.scss
```diff
@@ -46,3 +46,22 @@
 .uppercase {
   text-transform: uppercase;
 }
+
+.logo-doppler-new {
+  display: inline-block;
+  background-image: url('../img/doppler-logo-new.svg');
+  background-repeat: no-repeat;
+  height: 40px;
```
cbernat: Are you sure you need to restrict height and width at the same time? I think with one measure you'll do. Using both can lead to weird stretching of the image if sizes are not exact. I don't know if this is the case.

PR #65, opened 2019-03-26
Kpi reports

GusBaamonde:
- Two KPI states (Key Performance Indicator) were created.
- The preload for the boxes was created.
- The icons were added.
- A mixin was created to create the boxes.

https://cdn.fromdoppler.com/doppler-ui-library/build138/reports-filters.html


andresmoschini:
At a glance, I look it very nice, but I think that the right person to review it is @fcoronelmakingsense

assets/scss/modules/_box.scss
```diff
@@ -0,0 +1,10 @@
+@mixin dp-box ($color, $width){
+  border-radius: $dp-border-radius;
+  width: $width;
+  max-width: 408px;
```
damianmuti: Can we dynamically calculate this value based on how many columns the row of boxes should have? That would be really useful to have in this mixin.

assets/scss/modules/_box.scss
```diff
@@ -0,0 +1,10 @@
...
+  padding: 40px 30px;
+  overflow: hidden;
+  border:1px solid $color;
+  min-height: 395px;
```
damianmuti: Same goes with this value. ☝️

assets/scss/modules/_box.scss
```diff
@@ -0,0 +1,10 @@
+@mixin dp-box ($color, $width){
+  border-radius: $dp-border-radius;
+  width: $width;
+  max-width: 408px;
+  padding: 40px 30px;
```
damianmuti: Are we sure this boxes will have the same padding in handheld devices? If not, please use ems.

assets/scss/templates/_reportes.scss
```diff
@@ -77,3 +77,93 @@
     top: 43px;
   }
 }
+
+/* Section KPI */
+.container-kpi{
```
damianmuti: Missing empty space between classname and "{". Same goes with this whole block.

assets/templates/reports-filters.html
```diff
@@ -86,9 +86,39 @@ <h3>Optimiza tu estrategia de Marketing Digital</h3>
...
       </div><!-- hero-wrapper -->
     </header>
+    <section class="container-kpi">
+    <div class="wrapper-kpi">
+      <div class="reports-box warning--kpi">
+        <h3 class="number-kpi">0</h3>
+        <h6 class="subtitle-kpi">SUSCRIPTORES NUEVOS</h6>
+        <small class="date-range">12/02/2019 a 12/03/2019</small>
+        <p class="text-kpi">Suscriptores añadidos a través de tus automations por comportamiento en Sitio.</p>
+        <ul class="group-options">
```
damianmuti: This could be a `<nav>` tag. 👌

assets/scss/modules/_preload.scss
```diff
@@ -51,3 +51,45 @@
...
+  .loading-box {
+    position: relative;
+    display: flex;
+    top: 50%;
+    align-items: center;
+    justify-content: center;
+    color:$dp-color-lightgrey;
+
+    &:before {
+      @include dp-preloader($dp-color-lightgrey, 5em);
```
damianmuti: ❤️ 
One thing that comes to my mind looking at the deployed version, is that for this mixin we could move the spinner to the `::after` pseudo-element. Why?
By doing so, we could pass a new parameter to `dp-preloader` mixin with a hexa color value to use as a background that could cover the whole block in where this preloader is going to be used to truly hide the content until it gets hidden or removed from the DOM. We could create a `::before` pseudo-element (if the background parameter exists) in a fixed position to cover the container. 

Let's give it a thought. 👌

assets/scss/modules/_box.scss
GusBaamonde: ok I modify it

assets/scss/modules/_box.scss
GusBaamonde: ok I modify it

assets/scss/modules/_box.scss
GusBaamonde: ok i change it to ems

assets/scss/modules/_preload.scss
GusBaamonde: we could add it as an improvement, it's a good idea, I need to think it well

assets/scss/templates/_reportes.scss
```diff
@@ -77,3 +77,97 @@
...
+    a {
+      color:$dp-color-green;
+      font-weight: bold;
+      text-transform: uppercase;
+      font-size: $dp-font-size-xsmall;
+      font-family: Arial;
+      @include transition-all;
+    }
+    a:hover {
+      color:#008046;
```
cbernat: I think this color should be in a variable

GusBaamonde: Thanks @cbernat I change it

PR #69, opened 2019-03-28
feat: Added the html elements and login form styles

GusBaamonde:
- The html elements and form styles of the login were added. 
- The spaces will be normalized in another PR.
-  The header will be changed in another PR.

https://cdn.fromdoppler.com/doppler-ui-library/build155/login.html

### DIFF
--- a/assets/scss/modules/_forms.scss
+++ b/assets/scss/modules/_forms.scss
@@ -54,3 +54,29 @@
     transform: rotate(45deg);
   }
 }
+
+/* login form */
+
+.login-form {
+
+  .legend {
+    font-size: $dp-font-size-small;
+    float: right;
+    color: $dp-color-sylver;
+    font-style: italic;
+    font-weight: $dp-font-weight-light;
+  }
+
+  .forgot-link {
+    display: block;
+    margin: $dp-spaces--lv4 $dp-spaces--lv0;
+    color: $dp-color-green;
+    font-size: $dp-font-size-form;
+    font-style: italic;
+    font-family: $dp-font-family;
+
+      &:hover {
+        color: $dp-color-darkgreen;
+      }
+  }
+}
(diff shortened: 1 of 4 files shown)


assets/scss/modules/_forms.scss
```diff
@@ -54,3 +54,29 @@
...
+  .forgot-link {
+    display: block;
+    margin: $dp-spaces--lv4 $dp-spaces--lv0;
+    color: $dp-color-green;
+    font-size: $dp-font-size-form;
+    font-style: italic;
+    font-family: $dp-font-family;
+
+      &:hover {
+        color: $dp-color-darkgreen;
```
andresmoschini: Why do we need a special hover color for a particular link? (Probably the same question applies for most of the `.forgot-link`'s properties)

assets/scss/templates/_signup.scss
```diff
@@ -6,9 +6,9 @@
   align-items: stretch;
 
   .main-panel {
-    width: 500px;
+    width: 525px;
     height: 100vh;
-    padding: 40px 50px;
+    padding: 40px $dp-spaces--lv6;
```
andresmoschini: 40px is not standard? maybe we should normalize it.

andresmoschini:
Nice, very simple changes!

assets/scss/modules/_forms.scss
GusBaamonde: All links are green and dark green in hover.
Maybe we would have to change it in base.scss for all the links, and only add the classes for special cases.

assets/scss/templates/_signup.scss
GusBaamonde: I'm going to normalize it in another PR

andresmoschini: 👍

PR #70, opened 2019-03-28
fix: normalize spaces in login and signup

GusBaamonde:
- Normalize spaces in login and signup

https://cdn.fromdoppler.com/doppler-ui-library/build157/login.html
https://cdn.fromdoppler.com/doppler-ui-library/build157/signup.html


andresmoschini:
❤️

PR #75, opened 2019-04-03
fix: I deleted the phrase and put the show password icon

GusBaamonde:
I deleted the phrase in login and put the show password icon


andresmoschini:
👍

PR #76, opened 2019-04-04
fix: icon fonts not being generated

damianmuti:
Updated Gulp logic
Updated .gitignore


andresmoschini:
👏 It is fixed now! https://cdn.fromdoppler.com/doppler-ui-library/build174/signup.html

PR #81, opened 2019-04-08
Improvement: refactored language selector component

damianmuti:
I've refactored the whole language selector component, since the structure was complicated to maintain and to manipulate content inside each of the language buttons.
Now we have a cleaner solution with a list of elements that are exactly the same.

image


andresmoschini:
Great! thanks @damianmuti ❤️

andresmoschini:
:shipit:

assets/templates/signup.html
```diff
@@ -51,9 +51,15 @@
...
-            <div class="option" id="spanish-selector">
-              <span><img src="img/spanish-flag.svg"><strong>ES</strong></span>
-              <a class="option--item" href="#"><img src="img/english-flag.svg"><strong>EN</strong></a>
+            <div class="lang-option option--selector">
+              <button class="lang--es">
+                ES
+              </button>
+            </div>
+            <div class="lang-option">
+              <button class="lang--en">
```
fcoronelmakingsense: Need this in `<a>` because is a link

GusBaamonde:
Cuando despliega la banderita si clickeas fuera no se oculta y creo que @fcoronelmakingsense necesitaba que no sean buttons, creo que prefiere links

assets/templates/signup.html
andresmoschini: Good catch @fcoronelmakingsense! 

@damianmuti what do you think about fixing it and also applying the related changes in webapp before merging it?

damianmuti: Sure, @andresmoschini ! I've just pushed the changes to this PR 👌

andresmoschini: Fo, @fcoronelmakingsense, feel free to ignore these changes in your tasks related to signup page, forgot page, etc.

PR #82, opened 2019-04-09
chore: update packages

cbernat:
Updated all packages to latest version using https://www.npmjs.com/package/npm-check-updates
Tested gulp run, gulp build, and gulp dist tasks.


package.json
```diff
@@ -24,37 +24,39 @@
...
   },
   "main": "pdr",
   "engines": {
     "node": ">=6.9.1",
     "npm": ">=3.10.9"
   },
-  "dependencies": {}
+  "dependencies": {
+    "gulp": "^3.9.1",
+    "ttf2woff2": "^2.0.3"
```
andresmoschini: These are not devDependencies?

cbernat: Yes, you are right.

PR #84, opened 2019-04-10
fix: general adjustments of margins and paddings for 320px

GusBaamonde:
General adjustments of margins and paddings for 320px
Reviewed with Santiago Paolucci

320_layout

### DIFF
--- a/assets/scss/templates/_signup.scss
+++ b/assets/scss/templates/_signup.scss
@@ -123,7 +123,24 @@
     }
   }
 }
-// Language selector
+
+@include media-breakpoint('mobilexs') {
+
+  .panel-wrapper{
+    .main-panel {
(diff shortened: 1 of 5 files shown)


andresmoschini:
At a glance it looks fine, do you have a mockup in the special resolutions?

assets/scss/templates/_signup.scss
```diff
@@ -123,7 +123,24 @@
...
 }
-// Language selector
+
+@include media-breakpoint('mobilexs') {
+
+  .panel-wrapper{
+    .main-panel {
+        padding: $dp-spaces--lv5 $dp-spaces--lv4;
+
+      h5 {
```
cbernat: Could you please check tabulation here?, it seems a bit odd.

PR #85, opened 2019-04-11
Replacement of font size variables

GusBaamonde:
I removed the old font size variables, replacing them with new variables in levels, impacting all the scss files. I've talked to Santiago Paolucci and he agrees.

variables-font-size

### DIFF
--- a/assets/scss/core/_variables.scss
+++ b/assets/scss/core/_variables.scss
@@ -25,32 +25,6 @@ $dp-font-color-bright: #edede6 !default;
 $dp-font-color-dark: $dp-font-color !default;
 $dp-font-color-white: #fff !default;
 
-
-////
-/// Font Size settings
-///
-
-$dp-font-size--lv0: 0;
-$dp-font-size--lv1: 10px;
-$dp-font-size--lv2: 11px;
-$dp-font-size--lv3: 12px;
-$dp-font-size--lv4: 13px;
-$dp-font-size--lv5: 14px;
-$dp-font-size--lv6: 15px;
-$dp-font-size--lv7: 16px;
-$dp-font-size--lv8: 17px;
-$dp-font-size--lv9: 18px;
-$dp-font-size--lv10: 19px;
-$dp-font-size--lv11: 20px;
-$dp-font-size--lv12: 21px;
-$dp-font-size--lv13: 22px;
-$dp-font-size--lv14: 23px;
-$dp-font-size--lv15: 24px;
-$dp-font-size--lv16: 30px;
-$dp-font-size--lv17: 40px;
-$dp-font-size--lv18: 50px;
-$dp-font-size--lv19: 60px;
-
 //
 // variables margins/paddings
 //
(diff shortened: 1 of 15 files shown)


assets/scss/core/_variables.scss
```diff
@@ -25,26 +26,33 @@ $dp-font-color-dark: $dp-font-color !default;
...
-$dp-font-button-medium: 13px;
-$dp-font-button-small: 11px;
-$dp-font-size-15: 15px;
-$dp-font-size-24: 24px;
-$dp-font-small-lineheight: 18px;
+////
+/// Font Size settings
+///
+
+$dp-font-size--lv0: 0;
```
andresmoschini: Does this variable make sense?

andresmoschini:
It is fine for me, maybe @mvillaseco would like to review it.

assets/scss/core/_variables.scss
GusBaamonde: En algunos casos necesitamos ocultar el contenido, actualmente pasa con el logo en signup, el navegador lo lee pero no tiene tamaño el contenido y se coloca la imagen por background

andresmoschini: Thanks for the explanation, you are the expert.

assets/scss/core/_variables.scss
```diff
@@ -25,26 +26,33 @@ $dp-font-color-dark: $dp-font-color !default;
...
+$dp-font-size--lv1: 10px;
+$dp-font-size--lv2: 11px;
+$dp-font-size--lv3: 12px;
+$dp-font-size--lv4: 13px;
+$dp-font-size--lv5: 14px;
+$dp-font-size--lv6: 15px;
```
cbernat: Do we use all this fontsizes for real? Seems a bit too much

mvillaseco: Omg, 19 font sizes? are we using them as other things also?(maybe a rename should be done)

damianmuti:
Despite the fact that we all agreed to have levels for sizes, it seems that this approach lacks of any sort of description in the naming.
Just as an exercise for anyone that looks at this comment, try to think about what's the font size for buttons with this new naming convention (without scrolling up). No clue? Me neither. 😝

assets/scss/core/_variables.scss
andresmoschini: I agree as we have standardized spaces for common usages, we should have standardized font sizes for common usages.

IgnacioRodrigues:
This nomenclature of variables seems not to be the correct one, like knowing what level to use in each case? Support to what they say in which the nomenclatures tell you to use them, with this change you lose the sense of using classes.

GusBaamonde:
fonts


**To unify criteria of variables, I talked with Santiago Paolucci, @mvillaseco  and @damianmuti and I think we reached a midpoint.**

- I made a survey of the size fonts currently used.
- I reduced the quantity of size fonts used.
- We will use the guide already posed in the mockup of the library for future designs (we talk with Santiago Paolucci and compare with the fonts of the mockup of the library).
- We will use more global levels (We speak with @mvillaseco ).
- We will not have specific names at the moment. (We talked to @damianmuti , and it's a good idea to make 'variables of variables' in order to make themes, improve for the future)
- I changed and cleaned all the font size variables we are using.

andresmoschini:
Looks fine!

PR #86, opened 2019-04-16
feat: separation of buttons in modal form

GusBaamonde:
separation of buttons in modal form DBR 175

separate-buttons


assets/scss/modules/_forms.scss
```diff
@@ -13,6 +13,10 @@
     padding-top: 20px;
     border-top: 1px solid $dp-color-sylver;
     text-align: right;
+
+    .dp-button {
+      margin-left: 10px;
```
andresmoschini: Should we not use the spacing level variables?

PR #93, opened 2019-04-26
BREAKING CHANGE: show notification message

GusBaamonde:
- Adapt the user-menu container to be able to show it by class and not by ID, reusing the style of the container

panel-notificaciones

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -12,7 +12,7 @@ $("body").on({
 }, ".menu-main .submenu-item");
 
 $("body").on('click', '.user-menu--open', function() {
- $(this).parent().find('.user-menu').toggleClass('open');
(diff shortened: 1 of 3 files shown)


assets/js/main.js
```diff
@@ -11,8 +11,8 @@ $("body").on({
   }
 }, ".menu-main .submenu-item");
 
-$("body").on('click', '#user-menu--open', function() {
-  $('.user-menu').toggleClass('open');
+$("body").on('click', '.user-menu--open', function() {
```
andresmoschini: It is not a breaking change?

GusBaamonde: you're right

andresmoschini:
Relacionado con: https://github.com/FromDoppler/doppler-webapp/pull/148

PR #94, opened 2019-04-29
DBR-213 - General form messages

andresmoschini:
Hey team! I am touching SCSS hehe!

The idea is unifying general form errors and field errors, and in both of them support basic HTML: p, span, strong, em, links, etc. 

By the moment, li is not being supported, because I do not need it yet.

Could you review it?

Thanks @GusBaamonde for your help!

@mvillaseco, @cbernat, @IgnacioRodrigues, @damianmuti 

You can review it in (in the mockups the font is wrong, in build 224 it is fixed)

* https://cdn.fromdoppler.com/doppler-ui-library/build224/signup.html
    
    image

* https://cdn.fromdoppler.com/doppler-ui-library/build224/login.html

    image

### DIFF
--- a/assets/scss/core/_base.scss
+++ b/assets/scss/core/_base.scss
@@ -246,11 +246,19 @@ input[type="tel"] {
 
   .wrapper-errors {
     display: inline-block;
+    margin: $dp-spaces--lv0;
+    color: $dp-color-red;
+    font-size: $dp-sizing--lvl2;
 
-    .error-message {
-      margin: $dp-spaces--lv0;
+    a {
       color: $dp-color-red;
+      text-decoration: underline;
+    }
+
+    p {
+      margin: $dp-spaces--lv0;
       font-size: $dp-sizing--lvl2;
+      line-height: $dp-spaces--lv3;
     }
   }
   /* animation to show the errors (optional) */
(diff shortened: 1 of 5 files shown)


assets/scss/core/_base.scss
```diff
@@ -236,6 +236,11 @@ input[type="tel"] {
   display: none;
 }
 
+.form-message {
+  margin-bottom: $dp-spaces--lv2;
+  font-family: $dp-font-family;
```
andresmoschini: Hey @GusBaamonde, we did not see it, but without this change, span is rendered with serif font. Do you think that it is fine? should apply it in another place?

andresmoschini: @GusBaamonde commented me that @santiagopaolucci said that we should use `$dp-font-family-proximanova` in all public pages.

I will check if "validation errors" is the only missing, in that case, I will fix it, if not, I will create an issue.

andresmoschini:
You can review it in 

* https://cdn.fromdoppler.com/doppler-ui-library/build220/signup.html
    
    image

* https://cdn.fromdoppler.com/doppler-ui-library/build220/login.html

    image

andresmoschini:
> ⚠️ No mergear! Falta la manito en el link!

Listo, faltaba el href

assets/scss/core/_base.scss
andresmoschini: The fix is ready! Resolving the conversation...

cbernat:
Great job! All in order 👍

PR #96, opened 2019-04-30
Create container to display datahub messages

GusBaamonde:
Create container to display the datahub messages.
This container will allow us to continue implementing the designs of the messages.

PR realated with:  https://github.com/FromDoppler/doppler-webapp/pull/151

datahub-messages

### DIFF
--- a/assets/scss/core/_variables.scss
+++ b/assets/scss/core/_variables.scss
@@ -14,7 +14,7 @@ $dp-font-small-lineheight: 18px;
 $dp-font-weight: 400 !default;
 $dp-font-family: Arial, sans-serif !default;
 $dp-font-family-proximanova: "proxima-nova", Helvetica;
-$dp-font-size: 16px !default;
+$dp-font-size: 15px !default;
 $dp-font: $dp-font-weight $dp-font-size $dp-font-family !default;
 $dp-font-weight-light: 300 !default;
 $dp-font-weight-regular: $dp-font-weight !default;
--- a/assets/scss/templates/_reportes.scss
+++ b/assets/scss/templates/_reportes.scss
@@ -1,3 +1,27 @@
+
+.dp-datahub-messages {
+  max-width: $dp-wrapper-max-width;
+  margin: 0 auto;
+
+  h4 {
+    margin: $dp-spaces--lv0;
+    font-weight: $dp-font-weight-bold;
+    color: $dp-color-grey;
+  }
+
+  p {
+    margin: $dp-spaces--lv0 $dp-spaces--lv0 $dp-spaces--lv2;
+  }
+
+  .dp-datahub-action {
+    text-align: right;
+    margin-top: $dp-spaces--lv6;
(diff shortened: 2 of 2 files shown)


assets/scss/templates/_reportes.scss
```diff
@@ -1,3 +1,27 @@
+
+.dp-datahub-messages {
```
andresmoschini: What do you think about _dp-rich-message_?

assets/scss/templates/_reportes.scss
```diff
@@ -1,3 +1,27 @@
...
+    margin: $dp-spaces--lv0;
+    font-weight: $dp-font-weight-bold;
+    color: $dp-color-grey;
+  }
+
+  p {
+    margin: $dp-spaces--lv0 $dp-spaces--lv0 $dp-spaces--lv2;
+  }
+
+  .dp-datahub-action {
```
andresmoschini: Again, what do you think about remove datahub from here?

assets/scss/templates/_reportes.scss
```diff
@@ -81,7 +105,7 @@
 /* Section KPI */
 .container-reports {
   width: 100%;
-  padding: 60px 0;
+  padding: 60px 2%;
```
andresmoschini: percentage? is it fine?

Should you not use the levels?

assets/scss/core/_variables.scss
```diff
@@ -14,7 +14,7 @@ $dp-font-small-lineheight: 18px;
 $dp-font-weight: 400 !default;
 $dp-font-family: Arial, sans-serif !default;
 $dp-font-family-proximanova: "proxima-nova", Helvetica;
-$dp-font-size: 16px !default;
+$dp-font-size: 15px !default;
```
andresmoschini: Is there not a font size level?

assets/scss/templates/_reportes.scss
andresmoschini: I saw that you have changed it by levels, but, I am still curious, is it the same? are levels not absolute and percentage relative? how are they equivalent?

GusBaamonde: El porcentaje es relativo y los niveles son absolutos, solo tome algo proporcional. no es lo mismo. se uso porcentaje para que sea relativo al tamaño de pantalla. Pero como planteamos niveles no estaría mal utlizarlos.

assets/scss/core/_variables.scss
GusBaamonde: This case is an exception. The levels we have jump from 6 to 6. 12 is very small and 18 is too big. The mockups have the size in 15. I preferred to adjust to the design. In any case, this size is within a variable and can be modified.

andresmoschini: I is fine for me, thanks for the clarification!

andresmoschini:
I think that it is a nice step in the right direction. Thanks @GusBaamonde !

:shipit:

PR #105, opened 2019-05-13
fix: links should inherit font-size

damianmuti:
For the login, signup, and forgot password pages, register/signup links after the main header should inherit their font-size.


andresmoschini:
Did you verify how does it affect both _UI Library_ and _WebApp_?

damianmuti:
> Did you verify how does it affect both _UI Library_ and _WebApp_?

Yes. This class only affects links in headers in those pages.

PR #107, opened 2019-05-14
visual error in upgrade submit

GusBaamonde:
Flex properties are not well supported by fieldset elements.

fieldset-flex


andresmoschini:
👍 

@damianmuti could you review it?

PR #116, opened 2019-05-23
fix: reports rankings spinner style

fcoronelmakingsense:
- Fix spinner in reports rankings section to see inside the box

### DIFF
--- a/assets/scss/templates/_reportes.scss
+++ b/assets/scss/templates/_reportes.scss
@@ -225,6 +225,7 @@
   max-width: $dp-wrapper-max-width;
   margin: $dp-spaces--lv4 auto;
   font-family: $dp-font-family;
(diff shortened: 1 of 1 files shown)


cbernat:
👍

assets/scss/templates/_reportes.scss
```diff
@@ -225,6 +225,7 @@
   max-width: $dp-wrapper-max-width;
   margin: $dp-spaces--lv4 auto;
   font-family: $dp-font-family;
+  position: relative;
```
andresmoschini: Is that enough? does it not break anything?

fcoronelmakingsense: its necessary because the spinner has absolute position and need parent with relative to align inside the ranking box

PR #117, opened 2019-05-29
refactor: dropdowns now are accessible with keyboard

damianmuti:
One improvement that we've detected on the login/signup/forgot-password pages was that those pages are not keyboard accessible.
What we've ended up deciding to do was to refactor what we needed for keyboard accessibility to work, including dropdown menus.
Instead of creating something from scratch, I've checked what W3C says about dropdown menus and, to our fortune, they provide some javascript+html+css standalone modules for this to properly work, so this PR is based on that.

image
image
image
image

### DIFF
--- a/assets/js/MenuItemLinks.js
+++ b/assets/js/MenuItemLinks.js
@@ -1,0 +1,160 @@
+/*
+*   This content is licensed according to the W3C Software License at
+*   https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
+*
+*   File:   MenuItemLinks.js
+*
+*   Desc:   Popup Menu Menuitem widget that implements ARIA Authoring Practices
+*/
+
+/*
+*   @constructor MenuItemLinks
+*
+*   @desc
+*       Wrapper object for a simple menu item in a popup menu
+*
+*   @param domNode
+*       The DOM element node that serves as the menu item container.
+*       The menuObj PopupMenu is responsible for checking that it has
+*       requisite metadata, e.g. role="menuitem".
+*
+*   @param menuObj
+*       The object that is a wrapper for the PopupMenu DOM element that
+*       contains the menu item DOM element. See PopupMenu.js
+*/
+var MenuItemLinks = function (domNode, menuObj) {
+
+  this.domNode = domNode;
+  this.menu = menuObj;
+
+  this.keyCode = Object.freeze({
+    'TAB': 9,
+    'RETURN': 13,
+    'ESC': 27,
+    'SPACE': 32,
+    'PAGEUP': 33,
+    'PAGEDOWN': 34,
+    'END': 35,
+    'HOME': 36,
+    'LEFT': 37,
+    'UP': 38,
+    'RIGHT': 39,
+    'DOWN': 40
+  });
+};
+
+MenuItemLinks.prototype.init = function () {
+  this.domNode.tabIndex = -1;
+
+  if (!this.domNode.getAttribute('role')) {
+    this.domNode.setAttribute('role', 'menuitem');
+  }
+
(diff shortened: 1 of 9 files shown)


GusBaamonde:
Spectacular! :+1:

cbernat:
@damianmuti I don't seem to view this working, the url is this one right? https://cdn.fromdoppler.com/doppler-ui-library/build294/forgot-password.html

damianmuti:
> @damianmuti I don't seem to view this working, the url is this one right? https://cdn.fromdoppler.com/doppler-ui-library/build294/forgot-password.html

I've just pushed some changes since the build task seems not to include the menu instance inside the app.js file, @cbernat

IgnacioRodrigues:
Good work Dami, this improvement is necessary to apply the Doppler site.

assets/templates/forgot-password.html
```diff
@@ -103,6 +103,12 @@ <h1>Suma contactos a tus Listas con Formularios personalizados</h1>
   <!-- build:js:dist js/app.js -->
   <script src="./js/main.js"></script>
   <!-- /build -->
+  <script type="text/javascript">
+    window.onload=function() {
```
cbernat: This can't be avoided, can it?

damianmuti: The module needs a way to be instantiated; I guess its a good idea not to include the initialization in the library and make it whenever its needed instead, @cbernat

cbernat:
It's looking very nice! Good job @damianmuti ❤️

assets/js/MenuItemLinks.js
```diff
@@ -0,0 +1,160 @@
...
+*   @param menuObj
+*       The object that is a wrapper for the PopupMenu DOM element that
+*       contains the menu item DOM element. See PopupMenu.js
+*/
+var MenuItemLinks = function (domNode, menuObj) {
+
+  this.domNode = domNode;
+  this.menu = menuObj;
+
+  this.keyCode = Object.freeze({
```
andresmoschini: I am not sure how compatible is it with old browsers, maybe we should use simple objects.

damianmuti: @andresmoschini this is part of the component files that W3C provides for the component to work.

assets/js/MenuItemLinks.js
```diff
@@ -0,0 +1,160 @@
+/*
+*   This content is licensed according to the W3C Software License at
+*   https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
```
andresmoschini: Did you verify if we are following the license requirements?

assets/templates/forgot-password.html
andresmoschini: What do you think about making doing something to automatically detect these elements and apply the behavior?

andresmoschini:
@damianmuti creo que el comportamiento no es el mismo que antes. Antes, para abrir el menú había que hacer click, ahora alcanza con el hover.

Es discutible si eso es mejor o no, pero en este caso debemos mantener consistencia con el mismo menú de lenguaje que está en el sitio, por lo que creo que deberíamos tratar de respetar el comportamiento anterior.

Cualquier cosa avisame y lo hablamos.

assets/js/MenuItemLinks.js
andresmoschini: Si, lo sé, pero aún así no se que tan compatible es, y creo que ese cambio no hace a la funcionalidad, solo a las buenas prácticas.

andresmoschini: Com @damianmuti ya verificamos que _Object.freeze_ es muy compatible, no deberíamos preocuparnos.

PR #122, opened 2019-06-13
Message warning

GusBaamonde:
messages-warning


andresmoschini:
❤️

assets/scss/modules/_forms.scss
```diff
@@ -337,6 +337,10 @@ input[type="tel"] {
   background-repeat: no-repeat;
   font-family: $dp-font-family-proximanova;
 
+  a {
```
cbernat: Weren't we avoing general styles over elements? Just in case that we want to have another `a` without underline on this same form.

cbernat:
Very nice @GusBaamonde !

assets/scss/modules/_forms.scss
andresmoschini: It is `.form-message a`, I mean, it is not a style in a general element.

image

assets/scss/modules/_forms.scss
```diff
@@ -337,6 +337,10 @@ input[type="tel"] {
   background-repeat: no-repeat;
   font-family: $dp-font-family-proximanova;
 
+  button, a{
+    text-decoration: underline;
+  }
+
 /* Validations Unexpected message */
 &.dp-error {
```
cbernat: This is not part of this PR but if `&.dp-error` is inside `.form-message` the tabulation may be a bit off. The same with the other classes below.

assets/scss/modules/_forms.scss
cbernat: You are right! My mistake, thanks for the explanation.

damianmuti:
GJ Gus! 👍

PR #124, opened 2019-06-18
feat: change height logo client manager

GusBaamonde:
fix-logo-cm

### DIFF
--- a/assets/scss/templates/_header.scss
+++ b/assets/scss/templates/_header.scss
@@ -41,7 +41,7 @@
 
     img {
       display: block;
-      height: auto;
+      height: 50px;
     }
   }
   .logo {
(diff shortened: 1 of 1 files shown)


assets/scss/templates/_header.scss
```diff
@@ -41,7 +41,7 @@
 
     img {
       display: block;
-      height: auto;
+      height: 50px;
```
cbernat: Why the sudden hardcoded height? Is it related to the change in dopplerApp?

GusBaamonde: Yes, in a meeting with @CarlosSampedro  and @santiagopaolucci it was defined that the logo has an upper padding of 18px, without lower padding in the image and that the padding remains 27.5px that currently has the header in doppler

PR #126, opened 2019-06-19
fix: submenu DBR-339

GusBaamonde:
prod-menu

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -6,7 +6,7 @@ $('body').on({
       $('.header-main').removeClass('header-open');
     }
 
-    if (submenu && submenu[0].children.length) {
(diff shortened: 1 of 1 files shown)


assets/js/main.js
```diff
@@ -6,7 +6,7 @@ $('body').on({
       $('.header-main').removeClass('header-open');
     }
 
-    if (submenu && submenu[0].children.length) {
+    if (submenu[0] && submenu[0].children.length) {
```
cbernat: Generally in JQuery for validation of existence we use `.length` if it is 0 there is no element. and if it is more than 0 the element exists. It is working as you've solved as well, so it's ok too.

PR #127, opened 2019-06-21
feat: show notifications

GusBaamonde:
show-bell

Solved the :after data-count="0" @cbernat

### DIFF
--- a/assets/scss/templates/_header.scss
+++ b/assets/scss/templates/_header.scss
@@ -182,11 +182,12 @@
     color: $dp-color-silver;
   }
 
-  a.active {
+  .active {
     color: $dp-color-lightgrey;
+    cursor: pointer;
   }
 
-  a.active:after {
+  .active:after {
     content: attr(data-count);
     color: $dp-color-white;
     background-color: $dp-color-yellow;
@@ -197,8 +198,8 @@
     font-size: 11px;
     position: absolute;
     text-align: center;
-    top: -5px;
-    left: 10px;
+    top: 6px;
+    left: 9px;
   }
 
   .user-avatar {
--- a/assets/scss/templates/_header.scss
+++ b/assets/scss/templates/_header.scss
@@ -256,8 +256,8 @@
 }
 
 &.notifications {
-    padding: 25px;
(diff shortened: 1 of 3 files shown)


assets/scss/templates/_header.scss
```diff
@@ -251,10 +251,15 @@
...
   transition: all 0.3s ease;
 
-&.helper--right{
+&.helper--right {
   right: -12px;
 }
 
+&.notifications {
+    padding: 25px;
+    font-size: 13px;
```
cbernat: I think we have a variable for this font-size right?

assets/templates/header.html
```diff
@@ -69,10 +69,11 @@
           <ul class="nav-right-main--list">
             <li>
             <a href="#" class="user-menu--open active" data-count="1"><span class="ms-icon icon-notification"></span></a>
-            <div class="user-menu helper--right">
+            <div class="user-menu helper--right notifications">
               <h5>NUEVO: Comportamiento en Sitio</h5>
```
andresmoschini: It is missing changing this structure, right?

assets/scss/templates/_header.scss
```diff
@@ -256,8 +256,8 @@
 }
 
 &.notifications {
-    padding: 25px;
-    font-size: 13px;
+    padding: $dp-spaces--lv4;
+    font-size: $dp-sizing--lvl2;
```
cbernat: ❤️

cbernat:
I do not see free user notificiation here. The one with upgrade link.

assets/templates/header.html
GusBaamonde: Yes, just change it with the message structure that brings backend

GusBaamonde:
> I do not see free user notificiation here. The one with upgrade link.

El contenido del mensaje no viene por backend?

cbernat:
> > I do not see free user notificiation here. The one with upgrade link.
> 
> El contenido del mensaje no viene por backend?

Yes it does, but I thought we were implementing free message first, that's all, should be similar.

PR #128, opened 2019-06-25
feat: remove lateral margin from the menu

GusBaamonde:
dbr_342

Se observa una diferencia en el padding lateral de webapp en relación al resto de la app
http://jira.makingsense.com/browse/DBR-342

### DIFF
--- a/assets/scss/templates/_header.scss
+++ b/assets/scss/templates/_header.scss
@@ -31,7 +31,7 @@
 
 .dp-header--cm {
   .dp-logo--cm {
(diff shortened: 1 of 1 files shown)


cbernat:
👍

cbernat:
👍

andresmoschini:
If it does not have a breaking change, it could be merged without worry about webapp.

andresmoschini:
image

In my opinion, a change in a margin is not a feature, if before it was wrong, it is a fix. Could you rewrite the history fixing it?

PR #129, opened 2019-06-25
feat: remove overflow hidden from header main

GusBaamonde:
DBR-341 When header is collapsed, user lateral menu is hidden

shopy

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -8,11 +8,16 @@ $('body').on({
 
     if (submenu[0] && submenu[0].children.length) {
       $('.header-main').addClass('header-open');
+    } else {
+      $('.sub-menu.open').hide();
     }
   },
   mouseleave: function() {
-    if(!$(this).hasClass('submenu-item')) {
+    if ($('.header-main .submenu-item > a.active').length > 0){
       $('.header-main').addClass('header-open');
+      $('.sub-menu.open').show(150);
+    } else {
+      $('.header-main').removeClass('header-open');
     }
   }
 }, '.menu-main>li');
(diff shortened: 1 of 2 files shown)


assets/scss/templates/_header.scss
```diff
@@ -13,7 +13,6 @@
   display: flex;
   font-family: $dp-font-family;
   height: auto;
-  overflow: hidden;
```
andresmoschini: Does it not generate any unwanted behavior?

I mean, this CSS was there for a reason, right?

cbernat: For an unknown reason. This prevents the lateral menu for showing, I would say we need to test this thoroughly.

cbernat: Please check this case:
image

andresmoschini:
In my opinion, it is neither a feature. Could you rewrite the history fixing it?

Also, take care of not merging it before fixing the issue that @cbernat has found.

assets/js/main.js
```diff
@@ -8,11 +8,16 @@ $('body').on({
 
     if (submenu[0] && submenu[0].children.length) {
       $('.header-main').addClass('header-open');
+    } else {
+      $('.sub-menu.open').hide();
     }
   },
   mouseleave: function() {
-    if(!$(this).hasClass('submenu-item')) {
```
andresmoschini: in the new changes, _this_ is not used anymore, the affected element is not important anymore?

cbernat: I don't get what you mean. What new changes? Other than the ones in this PR?

cbernat: I use .header-main class that is the main class of the header. I suppose with this is more than ok.

PR #133, opened 2019-07-02
fix(header): add classes instead of styles to hide header submenu

cbernat:
Because react does not override inline styles there was an issue with the header.  
Scenario: when in reports, hover another menu without sub items and then go to shopify, the result was that the menu did not hide when in shopify. 

http://cdn.fromdoppler.com/doppler-ui-library/build391/header.html

### DIFF
--- a/assets/js/main.js
+++ b/assets/js/main.js
@@ -9,13 +9,13 @@ $('body').on({
     if (submenu[0] && submenu[0].children.length) {
       $('.header-main').addClass('header-open');
     } else {
-      $('.sub-menu.open').hide();
+      $('.sub-menu.open').removeClass('open');
     }
   },
   mouseleave: function() {
     if ($('.header-main .submenu-item > a.active').length > 0){
       $('.header-main').addClass('header-open');
-      $('.sub-menu.open').show(150);
+      $('.header-main .submenu-item > a.active + .sub-menu').addClass('open');
     } else {
       $('.header-main').removeClass('header-open');
     }


assets/scss/templates/_header.scss
```diff
@@ -123,6 +123,9 @@
 
   &.open {
     display: flex;
+    &.hidden {
```
fcoronelmakingsense: need hidden inside open?

assets/js/main.js
```diff
@@ -9,13 +9,13 @@ $('body').on({
...
     } else {
-      $('.sub-menu.open').hide();
+      $('.sub-menu.open').addClass('hidden');
     }
   },
   mouseleave: function() {
     if ($('.header-main .submenu-item > a.active').length > 0){
       $('.header-main').addClass('header-open');
-      $('.sub-menu.open').show(150);
+      $('.sub-menu.open').removeClass('hidden');
```
andresmoschini: 150 is not necessary anymore? is there a css animation now?

assets/scss/templates/_header.scss
cbernat: To hide menu when you do hover over templates menu for example. Perhaps open class should be named selected, or active.

assets/js/main.js
cbernat: I will add something here yes.

cbernat: I've tested this and don't see the white block we've seen previously. Seems that `removeClass` takes longer than `show`

cbernat: It's horrible as it was before, I think the way of fixing this is with animations. But this seems to be outside scope just yet.

andresmoschini: Makes sense, thanks for the update.

assets/scss/templates/_header.scss
cbernat: Moved class to atomic class stylesheet.

cbernat:
Test this please.

cbernat:
test this please

cbernat:
test this please

cbernat:
@andresmoschini @fcoronelmakingsense finally it was best to remove open class and add this class when needed. Tested in webapp and seems to be working fine. If you don't oppose I will be merging this today.

andresmoschini:
👍

PR #141, opened 2019-07-23
feat: Creation of tooltip component

GusBaamonde:
**Creation of tooltip component.**
To specify the orientation of the 'tooltip', you can use the top / right / bottom / left, classes if you need it.

tootltip-component

### DIFF
--- a/assets/scss/helpers/_colors.scss
+++ b/assets/scss/helpers/_colors.scss
@@ -20,6 +20,7 @@ $dp-color-lightred: #EC8A83;
 $dp-color-red: #E34F45;
 $dp-color-darkred: #A63A33;
 $dp-color-white: #FFFFFF;
+$dp-color-black: #000000;
 $dp-border-grey: #E8E8E8;
 $dp-color-default: #C9C7BD;
 $dp-color-lightbrown: #838370;
--- a/assets/scss/modules/_box.scss
+++ b/assets/scss/modules/_box.scss
@@ -4,7 +4,7 @@
 /// @access private
 /// @author gbaamonde@makingsense.com
 
-@mixin dp-box ($color:"", $width:"", $max-width:"", $min-height:"") {
+@mixin dp-box ($color:"$dp-color-silver", $width:"25%", $max-width:"400px", $min-height:"250px") {
   border-radius: $dp-border-radius;
   width: $width;
   max-width: $max-width;
--- a/assets/scss/modules/_box.scss
+++ b/assets/scss/modules/_box.scss
@@ -1,4 +1,10 @@
-@mixin dp-box ($color, $width, $max-width, $min-height){
+
+/// Box component
+/// We can use the following parameters to customize them $color, $width, $max-width, $min-height
+/// @access private
+/// @author gbaamonde@makingsense.com
+
+@mixin dp-box ($color, $width, $max-width, $min-height) {
   border-radius: $dp-border-radius;
   width: $width;
   max-width: $max-width;
--- a/assets/scss/modules/_box.scss
+++ b/assets/scss/modules/_box.scss
@@ -4,7 +4,7 @@
 /// @access private
 /// @author gbaamonde@makingsense.com
 
-@mixin dp-box ($color, $width, $max-width, $min-height) {
+@mixin dp-box ($color:"", $width:"", $max-width:"", $min-height:"") {
   border-radius: $dp-border-radius;
   width: $width;
   max-width: $max-width;
--- a/assets/scss/modules/_tooltip.scss
+++ b/assets/scss/modules/_tooltip.scss
@@ -10,7 +10,7 @@
   padding: 3px $dp-spaces--lv1;
   margin: 0;
   background: $bg-tooltip;
-  border-radius: 3px;
+  border-radius: $border-radius;
   font-size: $font-size-tooltip;
   text-transform: uppercase;
   color: $font-color-tooltip;
@@ -25,8 +25,7 @@
     background-color: $dp-color-black;
     color: $dp-color-white;
     text-align: center;
-    padding: 5px;
-    border-radius: 3px;
+    border-radius: $border-radius;
     z-index: 1;
     opacity: 0;
     transition: opacity .6s;
--- a/assets/scss/modules/_tooltip.scss
+++ b/assets/scss/modules/_tooltip.scss
@@ -2,7 +2,7 @@
 /// @example markup - HTML structure
 /// <p class="dp-tooltip-right">dooplertip <span class="tooltiptext">Tooltip text</span></p>
 /// In order to specify the orientation of the tooltip you can use the: dp-tooltip-top | dp-tooltip-right | dp-tooltip-bottom | dp-tooltip-left classes if you need it.
-
+/// @author gbaamonde@makingsense.com
 
 @mixin dp-tooltip ($bg-tooltip, $font-size-tooltip, $font-color-tooltip){
   position: relative;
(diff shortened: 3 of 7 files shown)


assets/scss/helpers/_colors.scss
```diff
@@ -20,6 +20,7 @@ $dp-color-lightred: #EC8A83;
 $dp-color-red: #E34F45;
 $dp-color-darkred: #A63A33;
 $dp-color-white: #FFFFFF;
+$dp-color-black: #000000;
```
fcoronelmakingsense: why use the long definition and don't the short?

assets/scss/modules/_tooltip.scss
```diff
@@ -1 +1,120 @@
...
+/// In order to specify the orientation of the tooltip you can use the: dp-tooltip-top | dp-tooltip-right | dp-tooltip-bottom | dp-tooltip-left classes if you need it.
 
+
+@mixin dp-tooltip ($bg-tooltip, $font-size-tooltip, $font-color-tooltip){
+  position: relative;
+  display: inline-block;
+  padding: 3px $dp-spaces--lv1;
+  margin: 0;
+  background: $bg-tooltip;
+  border-radius: 3px;
```
fcoronelmakingsense: maybe this value repeat in more places, can add in a variable?

assets/scss/modules/_tooltip.scss
```diff
@@ -1 +1,120 @@
...
+  .tooltiptext {
+    top: -50%;
+    right: 110%;
+
+    &:after {
+      top: 50%;
+      left: 100%;
+      margin-top: -5px;
+      border-color: transparent transparent transparent $dp-color-black;
+    }
```
fcoronelmakingsense: i don't know if posible extract these values in one function to reuse

assets/scss/templates/_guidelines.scss
```diff
@@ -209,3 +209,23 @@ body {
...
+
+.wrapper-tooltip{
+  flex-wrap: wrap;
+  display: flex;
+  justify-content: space-between;
+  padding: 100px 10px;
+}
+
+.dp-box{
+  @include dp-box($dp-color-silver, 30%, 408px, 250px);
```
fcoronelmakingsense: i dont know if better add a comment to know what means `408px` and `250px` because i dont understand what means

andresmoschini:
Buenos reviews @*fcoronelmakingsense !*

assets/scss/helpers/_colors.scss
GusBaamonde: We use the full color variables as there are more complex colors '# A63A33'

assets/scss/modules/_tooltip.scss
GusBaamonde: I can add the variable

assets/scss/modules/_tooltip.scss
GusBaamonde: In this case I don't think it's worth putting together a function, it's only 4 positions and I don't think we will use it in more places.

assets/scss/templates/_guidelines.scss
GusBaamonde: In this case it is a mix that we reuse from another component. Perhaps I could make the comments within that mix so that it remains in the documentation and better explain the parameters received.

assets/scss/modules/_tooltip.scss
```diff
@@ -1 +1,120 @@
...
+  text-transform: uppercase;
+  color: $font-color-tooltip;
+
+  .tooltiptext {
+    visibility: hidden;
+    position: absolute;
+    width: auto;
+    min-width: 170px;
+    display: inline-block;
+    padding: 5px;
```
cbernat: Repeated `padding: 5px;` please check. ln 28

assets/scss/modules/_tooltip.scss
```diff
@@ -1 +1,120 @@
...
+  margin: 0;
+  background: $bg-tooltip;
+  border-radius: 3px;
+  font-size: $font-size-tooltip;
+  text-transform: uppercase;
+  color: $font-color-tooltip;
+
+  .tooltiptext {
+    visibility: hidden;
+    position: absolute;
```
cbernat: Isn't there a better way than to work this out? (without position absolute here) perhaps working with flex-box?

assets/scss/modules/_tooltip.scss
```diff
@@ -1 +1,120 @@
+/// Tooltips Component
+/// @example markup - HTML structure
+/// <p class="dp-tooltip-right">dooplertip <span class="tooltiptext">Tooltip text</span></p>
+/// In order to specify the orientation of the tooltip you can use the: dp-tooltip-top | dp-tooltip-right | dp-tooltip-bottom | dp-tooltip-left classes if you need it.
 
+
+@mixin dp-tooltip ($bg-tooltip, $font-size-tooltip, $font-color-tooltip){
```
cbernat: Very nice! ❤️

assets/scss/templates/_guidelines.scss
fcoronelmakingsense: thanks!

assets/scss/modules/_tooltip.scss
GusBaamonde: Unfortunately we need the positioning to be able to control this. We cannot manipulate it only with flex.

assets/scss/templates/_guidelines.scss
```diff
@@ -209,3 +209,24 @@ body {
...
+  display: flex;
+  justify-content: space-between;
+  padding: 100px 10px;
+}
+
+.dp-box {
+  /* We can use the following parameters to customize them */
+  /* Mixin parameters $color - $width - $max-width - $min-height */
+
+  @include dp-box($dp-color-silver, 20%, 408px, 150px);
```
andresmoschini: Maybe this way is better to explain what is each parameter:

```suggestion
  @include dp-box($color: $dp-color-silver, $width: 20%, $max-width: 408px, $min-height: 150px);
```

See https://sass-lang.com/documentation/at-rules/mixin#keyword-arguments

Does it work?

GusBaamonde: I turned this question around and found a way to do it. I had to modify it in the creation of @mixin and not in the call of @include, sorry for the delay, I think it's more readable, thanks

assets/scss/modules/_box.scss
```diff
@@ -1,4 +1,10 @@
-@mixin dp-box ($color, $width, $max-width, $min-height){
+
+/// Box component
+/// We can use the following parameters to customize them $color, $width, $max-width, $min-height
+/// @access private
+/// @author gbaamonde@makingsense.com
+
+@mixin dp-box ($color:"", $width:"", $max-width:"", $min-height:"") {
```
andresmoschini: What do you think about using common default values?

assets/scss/templates/_guidelines.scss
```diff
@@ -209,3 +209,23 @@ body {
...
+.wrapper-tooltip {
+  flex-wrap: wrap;
+  display: flex;
+  justify-content: space-between;
+  padding: 100px 10px;
+}
+
+.dp-box {
+  /* We can use the following parameters to customize them */
+  /* Mixin parameters $color - $width - $max-width - $min-height */
```
andresmoschini: I think that the comments are not necessary anymore, what do you think @fcoronelmakingsense ?

assets/scss/modules/_box.scss
GusBaamonde: You're right, now I add the default values

PR #144, opened 2019-07-26
fix(table component)

GusBaamonde:
The separation line in the header of the table that was within `<a>` was removed, the separation line now depends on each `<th>`, so we avoid collecting items that are only decorative.

Ready @cbernat ! :)


cbernat:
Great! 👍

PR #146, opened 2019-07-29
fix(reports traffic sources): box height style

fcoronelmakingsense:
image


cbernat:
This goes with automerge @fcoronelmakingsense :) Green light!

PR #153, opened 2019-07-30
Fix component breadcrumb

GusBaamonde:
Remove styles for specific items

### DIFF
--- a/assets/scss/modules/_breadcrumb.scss
+++ b/assets/scss/modules/_breadcrumb.scss
@@ -39,6 +39,7 @@
 
     a {
       padding-left: 5px;
+      color: $dp-color-lightgrey;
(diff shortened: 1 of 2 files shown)


assets/scss/modules/_breadcrumb.scss
```diff
@@ -33,12 +33,13 @@
...
     color: $dp-color-lightgrey;
 
-    :before {
+    &:before {
       margin-right: 5px;
     }
 
     a {
       padding-left: 5px;
+      color: $dp-color-lightgrey;
```
andresmoschini: why do you need a special style for the `a` elements inside active breadcrumb item?

andresmoschini:
❤️

PR #154, opened 2019-07-31
fix: Include anchors in messages

GusBaamonde:
Include colors according to the message
Remove margin 0 in the `p`

links_in_messages


cbernat:
Looking good! 😄

PR #174, opened 2019-09-03
feat(partials): Use of partial for header, footer and aside

GusBaamonde:
-Add the links to the partials
-Add prettier for js files

partials


cbernat:
@GusBaamonde this is very nice, but perhaps you could divide into commits: one would be the prettier change and then partial handling, so it gets clearer to review? let me know if you need help with this.

GusBaamonde:
test this please

GusBaamonde:
test this please

andresmoschini:
test this please

andresmoschini:
test this please

assets/js/main.js
```diff
@@ -78,3 +78,23 @@ $(".show-hide").click(function() {
     $(this).removeClass("ms-icon icon-hide").addClass("ms-icon icon-view");
   }
 });
+
+// navigation partials
+
+$('.dpsg-link').on('click', function() {
+  var partial = $(this).attr('data-partial'),
+    elementClicked = this;
+  $.ajax({
```
cbernat: What happens if I click one tab, and then another, and then another, very fast? Perhaps you should ensure only one call happens at each time.

assets/js/main.js
```diff
@@ -78,3 +78,23 @@ $(".show-hide").click(function() {
     $(this).removeClass("ms-icon icon-hide").addClass("ms-icon icon-view");
   }
 });
+
+// navigation partials
+
+$('.dpsg-link').on('click', function() {
+  var partial = $(this).attr('data-partial'),
+    elementClicked = this;
```
cbernat: Wouldn't it be best to avoid having an extra variable here? `elementClicked` You could use `this` below

assets/js/main.js
```diff
@@ -78,3 +78,23 @@ $(".show-hide").click(function() {
...
+  $.ajax({
+    url: './partials/_' + partial + '.html',
+    type: 'GET',
+    dataType: 'html',
+    success: function(response) {
+      $('.dpsg-link').removeClass('active');
```
cbernat: What do you think about doing `$('.dpsg-link', this).toggleClass('active');` instead of 
```
$('.dpsg-link').removeClass('active');
$(elementClicked).addClass('active');
```

assets/scss/templates/_guidelines.scss
```diff
@@ -338,6 +338,7 @@ body {
   display: flex;
   position: relative;
   background: $dp-color-white;
+  /*border: 1px solid #e73c7e;*/
```
andresmoschini: Please remove commented code. Or, at least, add an explanatory comment.

cbernat:
Very nice! How about changing the PR name? you are not using ajax anymore.

GusBaamonde:
test this please

PR #176, opened 2019-09-09
feat(navigation): the @for was modified to take the active menu

GusBaamonde:
- the `@for` was modified to take the active menu according to the page
- The `data-name` of the html was modified to coordinate them with the menu


cbernat:
Very nice!

PR #177, opened 2019-09-11
feat(colors-component)

GusBaamonde:
- The color layout was created
- Color classes were created for reuse

color-component


assets/templates/colors-component.html
```diff
@@ -69,7 +69,314 @@
...
+            <div class="dpsg-content-colors">
+              <div class="dpsg-color-card dpsg-white">
+                <div class="dpsg-icon-color">Color</div>
+                <div class="dpsg-content-color">
+                  <h3>white</h3>
+                  <p class="dpsg-code">
+                    Color: #ffffff;
+                  </p>
+                  <p class="dpsg-code">
+                    .dp-white
```
andresmoschini: Each time that we add a new color, we should remember to update this page. 

Sadly, I think that it is very fragile, and it will very easy having inconsistencies in the future.

PR #181, opened 2019-09-16
feat(tooltip component): tooltip component adjustment

GusBaamonde:
- add tooltip svg icon
- modify the tooltip mixin

#Checklist

tooltip-component


assets/scss/modules/_tooltip.scss
```diff
@@ -5,31 +5,29 @@
...
+  cursor: pointer;
 
   .tooltiptext {
     visibility: hidden;
     position: absolute;
-    width: auto;
     min-width: 170px;
     display: inline-block;
-    padding: 5px;
+    padding: 14px 5px;
```
cbernat: Wouldn't be best here to have a padding of 15px instead of 14? to go from 5 to 5 and be able to reuse. What do you think?

GusBaamonde: Yes, you're right

PR #183, opened 2019-09-25
feat(tooltip): creation of a new tooltip 

GusBaamonde:
component for report charts visits weekdays hours

- Add a mixin for the new tooltip
- Add a different animation to the new tooltip

#Checklist

new-tooltip

### DIFF
--- a/assets/scss/modules/_tooltip.scss
+++ b/assets/scss/modules/_tooltip.scss
@@ -123,17 +123,18 @@
   .tooltip {
     display: block;
     position: relative;
-    top: -50px;
+    top: -30px;
     left: 50%;
     transform: translate(-50%, -50%) scale(0.75) rotate(15deg);
     transform-origin: bottom center;
-    padding: $dp-spaces--lv3 $dp-spaces--lv4;
+    padding: $dp-spaces--lv3;
     border-radius: 5px;
     background: rgba(0, 0, 0, 0.75);
     text-align: center;
     transition: 0.15s ease-in-out;
     opacity: 0;
     width: 100%;
+    min-width: 160px;
     pointer-events: none;
     z-index: 5;
     text-align: left;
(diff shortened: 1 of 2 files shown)


assets/scss/modules/_tooltip.scss
```diff
@@ -116,3 +116,82 @@
     }
   }
 }
+
+/* New Tooltip II */
```
fcoronelmakingsense: Maybe can use a name more descriptive

assets/scss/modules/_tooltip.scss
```diff
@@ -116,3 +116,82 @@
...
+    z-index: 5;
+    text-align: left;
+    border: 1px solid $dp-color-silver;
+    box-shadow: 0 0 0 4px rgba(204, 204, 204, 0.3);
+
+    &.bgColor {
+      background: rgba($dp-color-white, 0.95);
+    }
+
+    &:before {
```
fcoronelmakingsense: this is the arrow?

assets/scss/modules/_tooltip.scss
```diff
@@ -116,3 +116,82 @@
...
+      height: $dp-spaces--lv0;
+      border: 5px solid transparent;
+      position: absolute;
+      bottom: $dp-spaces--lv0;
+      left: 50%;
+      transform: translate(-50%, 100%);
+      border-top: 5px solid rgba($dp-color-white, 0.95);
+    }
+
+    p {
```
fcoronelmakingsense: it's ok the styles of the content inside the tooltip are here? in that case always `p, span` has this styles

assets/templates/tooltip-component.html
```diff
@@ -123,6 +123,27 @@ <h3>Tooltip</h3>
...
+              y horas de la pagina de reportes.
+            </p>
+            <hr class="dpsg-break" />
+            <div class="wrapper-tooltip">
+              <div class="dp-box">
+                <div class="dp-tooltip">
+                  <p class="wrap-tooltip dpsg-button-example">
+                    Tooltip II example
+                  </p>
+                  <div class="tooltip bgColor">
```
fcoronelmakingsense: only need in webapp this css class? or need all structure?

andresmoschini:
Nice! Does the tooltip work with CSS only? or the javascript was already there?

assets/scss/modules/_tooltip.scss
GusBaamonde: For me they are fine, because these elements have particular characteristics of this component

GusBaamonde:
> ¡Agradable! ¿La información sobre herramientas funciona solo con CSS? o el javascript ya estaba allí?

Only CSS

assets/scss/modules/_tooltip.scss
GusBaamonde: Yes

assets/templates/tooltip-component.html
GusBaamonde: the structure is practically similar, only using the classes should it work

PR #188, opened 2019-10-08
fix: kpi width adjustment

GusBaamonde:
#Checklist

ajuste-kpi


andresmoschini:
I though that this solution would look like a patch. But, in my opinion, this looks very nice!

PR #199, opened 2019-10-28
fix: decouple report styles

GusBaamonde:
#Checklist

### DIFF
--- a/assets/scss/templates/_reportes.scss
+++ b/assets/scss/templates/_reportes.scss
@@ -7,7 +7,7 @@
   box-shadow: 0 0 0 4px $dp-color-white;
   padding: $dp-spaces--lv4;
   margin: $dp-spaces--lv0 auto;
-  box-shadow: 0 0 0 4px #fff; /* TODO: check to modularize the messages */
+  box-shadow: 0 0 0 4px $dp-color-white; /* TODO: check to modularize the messages */
 
   h2 {
     color: $dp-color-darkgrey;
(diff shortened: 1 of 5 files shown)


assets/scss/templates/_reportes.scss
```diff
@@ -173,6 +173,35 @@
   }
 }
 
+/*********/
```
andresmoschini: What is the meaning of this separator?

assets/scss/templates/_reportes.scss
```diff
@@ -240,8 +269,8 @@
 
 /* Top ranking */
 .wrapper-reports-box {
-  max-width: $dp-wrapper-max-width;
-  margin: $dp-spaces--lv4 auto;
+  /*max-width: $dp-wrapper-max-width;*/
+  /*margin: $dp-spaces--lv4 auto;*/
```
andresmoschini: What do you think about removing commented code?

GusBaamonde: If I was just generating the version to be able to do tests on the webapp.
Then I delete everything and the word neat

assets/scss/templates/_reportes.scss
GusBaamonde: If I was just generating the version to be able to do tests on the webapp.
Then I delete everything and the word neat

assets/scss/templates/_reportes.scss
```diff
@@ -173,6 +173,35 @@
...
 
+/*********/
+
+.box-border--bottom {
+  position: relative;
+
+  &::after {
+    content: '';
+    position: absolute;
+    border-bottom: 3px solid #33ad73;
```
cbernat: You have a rogue color here `#33ad73`

PR #203, opened 2019-11-08
fix(dp-box): remove the overflow-hidden in the box-shadow mixin

GusBaamonde:
#Checklist

remove-overflow-hidden

### DIFF
--- a/assets/scss/modules/_box.scss
+++ b/assets/scss/modules/_box.scss
@@ -15,5 +15,4 @@
   min-height: $min-height;
   border: 1px solid $color;
   box-shadow: 0 0 0 4px rgba($color, 0.3);
-  overflow: hidden;
 }


andresmoschini:
❤️  removing code is great!

cbernat:
@GusBaamonde  do you remember why was thi overflow hidden in the first place?

GusBaamonde:
> @GusBaamonde , ¿recuerdas por qué se ocultó este desbordamiento en primer lugar?

If the overflow-hidden was placed so that nothing escapes from the boxes, but just with this graph the tooltip is cut, then it is necessary to remove it

andresmoschini:
> > @GusBaamonde , ¿recuerdas por qué se ocultó este desbordamiento en primer lugar?
> 
> If the overflow-hidden was placed so that nothing escapes from the boxes, but just with this graph the tooltip is cut, then it is necessary to remove it

This teaches us to avoid doing things _just in case_ :)

PR #219, opened 2019-12-04
fix: I removed shopify styles to reuse other webapp classes

GusBaamonde:
#Checklist

https://cdn.fromdoppler.com/doppler-ui-library/build671/integrations.html

layout-shopify

Related to: https://github.com/FromDoppler/doppler-webapp/pull/628

### DIFF
--- a/assets/scss/templates/_integrations.scss
+++ b/assets/scss/templates/_integrations.scss
@@ -2,51 +2,11 @@
   margin-top: 20px;
 
   @at-root {
-    &__block {
-      @include dp-box($dp-color-silver, 100%, initial, initial);
-
-      padding: 0;
-
-      hr {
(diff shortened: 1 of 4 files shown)


assets/scss/templates/_integrations.scss
```diff
@@ -2,7 +2,7 @@
   margin-top: 20px;
 
   @at-root {
-    &__block {
+    /* &__block {
```
andresmoschini: What do you think about directly removing this code in place of commented it?

GusBaamonde: Yes, at the moment I am doing tests to see it on the webapp, I promise to delete everything that is not used

cbernat:
👍

cbernat:
🚀

PR #226, opened 2019-12-27
feat(subscriber history): add icons buttons and validations

GusBaamonde:
#Checklist

form-subscribers-history


cbernat:
One question, right now we have only the first input, this does scale right? I mean, the search input streatchs to fill all width when the other three fields are not present?

GusBaamonde:
> Una pregunta, en este momento solo tenemos la primera entrada, ¿esto escala bien? Quiero decir, ¿la búsqueda ingresa streatchs para llenar todo el ancho cuando los otros tres campos no están presentes?

If perfect scale because we use the grid system to position the fields we need

andresmoschini:
👍

cbernat:
Very nice! ❤️

GusBaamonde:
SH-accesibilidad-01

assets/scss/modules/_images.scss
```diff
@@ -23,54 +22,63 @@
...
 // User removed soft bounces
-.icon-user--rsb:before {
+.user-removed-soft-bounces:before {
   @include icon-user-table;
   background-image: url('../../img/user-removed-soft-bounces.svg');
 }
 
 // User subscriber removed
-.icon-user--sr:before {
+.user-subscriber-removed:before {
```
cbernat: How about changing the order here to follow consistency? like `removed-subscriber`

assets/scss/modules/_images.scss
```diff
@@ -23,54 +22,63 @@
...
 // User active
-.icon-user--active:before {
+.user-active:before {
   @include icon-user-table;
   background-image: url('../../img/user-active.svg');
 }
 
 // User active not associated to lists
-.icon-user--natl:before {
+.user-active-with-no-list:before {
```
cbernat: Are we not using BEM here for class naming? I think here perhaps `user--active-with-no-list` would be best to represent element with modifier. What do you think?

GusBaamonde:
cambio-clases

PR #228, opened 2020-01-08
feat: add the table and campaign icons

GusBaamonde:
#Checklist

add-table-and-icons

### DIFF
--- a/assets/scss/modules/_images.scss
+++ b/assets/scss/modules/_images.scss
@@ -41,7 +41,8 @@
 %states-campaing {
   padding-left: $dp-spaces--lv3;
   display: flex;
-  align-items: center;
+  align-items: flex-start;
+  max-width: 200px;
 }
 
 .dp-calification {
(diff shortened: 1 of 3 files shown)


assets/scss/modules/_images.scss
```diff
@@ -22,6 +22,20 @@
   background-repeat: no-repeat;
 }
 
+///
+/// Mixin campaign status icons
+///
+
+@mixin icon-campaing($background-color) {
```
cbernat: ❤️

cbernat:
Did you try an example with very very long subject? For now I think perhaps showing ... would be enough. Or perhaps a line break. Check here https://projects.invisionapp.com/share/GPU9MUS5WY4#/screens/387473062 when campaign name is too long it breaks in more than one line.

GusBaamonde:
> Did you try an example with very very long subject? For now I think perhaps showing ... would be enough. Or perhaps a line break. Check here https://projects.invisionapp.com/share/GPU9MUS5WY4#/screens/387473062 when campaign name is too long it breaks in more than one line.

text-long

GusBaamonde:
ajuste-contenido

PR #230, opened 2020-01-10
fix: all subscriber status icons removed

GusBaamonde:
- 1 svg icon was implemented as font
- user status classes were respected to use the icon color
- the legend was added to each icon
- an arial-label and title were added to each icon

#Checklist

new-icons


cbernat:
How about adding the description text in the same color of the icon?

GusBaamonde:
> How about adding the description text in the same color of the icon?

We already have many problems of contrast with the design and problems of font sizes for everything that is Accessibility ... I would not color the typography, I think it would negatively impact the readability of each state

cbernat:
👍

PR #240, opened 2020-01-22
fix: the new background image

GusBaamonde:
the new background image was added and the image position was corrected

#Checklist

image-confirmation


cbernat:
👍

PR #246, opened 2020-02-03
fix: the pagination in the 980px resolution changes to the left side …

GusBaamonde:
the pagination in the 980px resolution changes to the left side of the grid

- add a right margin to the icons
- the tooltip is removed in resolution 1365px

#Checklist


cbernat:
👍

PR #278, opened 2020-03-27
fix: run prettier check on files

GusBaamonde:
#Checklist


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:10 as source
+#FROM node:10 as source
```
andresmoschini: It will not work

GusBaamonde: thanks!!

PR #284, opened 2020-04-02
fix: replace link with a button

GusBaamonde:
confirmation-page

#Checklist

### DIFF
--- a/assets/templates/confirmation-page.html
+++ b/assets/templates/confirmation-page.html
@@ -93,7 +93,7 @@
         </article>
         <p>
           ¿No has recibido el Email?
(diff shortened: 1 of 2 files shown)


assets/templates/confirmation-page.html
```diff
@@ -93,7 +93,7 @@ <h1>Gracias por registrarte</h1>
         </article>
         <p>
           ¿No has recibido el Email?
-          <a href="#" rel="noopener" target="_blank">Reenvíalo.</a>
+          <button type="button">Reenvíalo.</button>
```
andresmoschini, line 96: ❤️

PR #289, opened 2020-04-14
fix(semantic-release): fixed script

cbernat:
### Fix for release script
Semantic release needs dist folder created with build to generate the tag, this is a bit different in doppler webapp. 

#Checklist

@andresmoschini I'm tagging you here just in case.

### DIFF
--- a/build-n-release-w-docker.sh
+++ b/build-n-release-w-docker.sh
@@ -23,17 +23,20 @@ export MSYS2_ARG_CONV_EXCL="*"
 
 mkdir -p dist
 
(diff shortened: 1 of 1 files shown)


build-n-release-w-docker.sh
```diff
@@ -23,17 +23,20 @@ export MSYS2_ARG_CONV_EXCL="*"
 
 mkdir -p dist
 
-docker build --pull -t doppler-ui-system-source --target test .
+docker build --pull -t doppler-ui-system-source .
```
andresmoschini, line 26: I think that `--target test` is better because you are not taking advantage of the build

PR #291, opened 2020-04-15
fix: force new version

cbernat:
#Checklist


cbernat:
Removed the tag manually and now it's working as it should https://cdn.fromdoppler.com/doppler-ui-library/v3.45.2/notification-poc.html 
https://cdn.fromdoppler.com/doppler-ui-library/latest/notification-poc.html

PR #300, opened 2020-04-27
Dw 305 tabs component

GusBaamonde:
- add the html tabs file
- add the style for the component
- add the link in the navigation bar

tabs-component

#Checklist

### DIFF
--- a/assets/js/webPushMain.js
+++ b/assets/js/webPushMain.js
@@ -29,7 +29,6 @@ var UI = {
   },
 };
 
-if (!!UI.notificationButton) {
   UI.notificationButton.addEventListener('click', function () {
     if (isSubscribed) {
       sendRequestNotification('http://localhost:5000');
@@ -37,9 +36,7 @@ if (!!UI.notificationButton) {
       console.log('Web push: No esta suscripto a notificaciones');
     }
   });
-}
 
-if (!!UI.notificationButtonInt) {
   UI.notificationButtonInt.addEventListener('click', function () {
     if (isSubscribed) {
       sendRequestNotification('http://appint.fromdoppler.net:5001');
@@ -47,7 +44,6 @@ if (!!UI.notificationButtonInt) {
       console.log('Web push: No esta suscripto a notificaciones');
     }
   });
-}
 
 var isSubscribed = false;
 var swRegistration = null;
--- a/assets/js/webPushMain.js
+++ b/assets/js/webPushMain.js
@@ -1,3 +1,5 @@
+
+if (window.location.pathname === '/notification-poc.html') {
 var applicationServerPublicKey =
   'BFtBVunx6Lg_4ziA6b_Oe-9iqQetudiBzkukb1UOOWItZlbdPXCnMLjTfM6gnbmzrusaGlifwWWaFRWnef_H40Y';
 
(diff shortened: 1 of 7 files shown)


mdirago:
Excellent Gus!

assets/js/webPushMain.js
```diff
@@ -1,3 +1,5 @@
+
+if (window.location.pathname === '/notification-poc.html') {
```
cbernat, line 2: ❤️

assets/js/webPushMain.js
```diff
@@ -26,21 +28,26 @@ var UI = {
...
-});
 
-UI.notificationButtonInt.addEventListener('click', function () {
-  if (isSubscribed) {
-    sendRequestNotification('http://appint.fromdoppler.net:5001');
-  } else {
-    console.log('Web push: No esta suscripto a notificaciones');
-  }
-});
+if (!!UI.notificationButton) {
```
cbernat, line 32: I think you don't need this anymore 😄

assets/scss/modules/_tabs.scss
```diff
@@ -0,0 +1,104 @@
...
+
+.wrapper-tabs {
+  margin: 10px $dp-spaces--lv0 $dp-spaces--lv9;
+  position: relative;
+  width: 100%;
+}
+
+.bg-tabs {
+  position: absolute;
+  border-top: 1px solid #ccc;
```
cbernat, line 19: Please chech to use color variables here for `#ccc` and `#eaeaea`

assets/scss/templates/_guidelines.scss
```diff
@@ -212,7 +145,7 @@ body {
 .dpsg-main-wrap {
   display: flex;
   flex-direction: column;
-  font-family: $dp-font-family-proximanova;
+  font-family: Arial, sans-serif;
```
cbernat, line 148: How about adding these font types as a variable too? maybe we reuse them somewhere else.

cbernat:
Very nice! https://cdn.fromdoppler.com/doppler-ui-library/build856/tabs-component.html

andresmoschini:
Muy bueno!

assets/templates/tabs-component.html
```diff
@@ -0,0 +1,749 @@
...
+            <p>
+              Los tabs nos van a permitir crear una interfaz de usuario basada
+              en pestañas, donde, de una forma muy intuitiva, podemos ofrecer al
+              usuario diferentes vistas, que son seleccionadas al pulsar una de
+              las pestañas que se muestran a continuación.
+            </p>
+            <hr class="dpsg-break" />
+            <div class="dp-container">
+              <div class="dp-rowflex">
+                <nav class="wrapper-tabs">
```
cbernat, line 86: Perhaps here you could think of a more cohesive way for the naming of the classes, for example:
``` 
<nav class="tab-menu--wrapper">
      <ul class="tab-menu">
         <li class="tab-menu--item">
...
```
It gives you an idea of this tab as a component, tab-menu, which has many parts: a wrapper and items. 
I'm not so sure about `tab-menu`, perhaps another name would be ok, what do you think?

assets/scss/templates/_guidelines.scss
GusBaamonde: _$ dp-font-family: Arial, sans-serif! default;
$ dp-font-family-proximanova: 'proxima-nova', Helvetica, Arial, sans-serif! default;
$ dp-font-family-handofsean: 'hand-of-sean';_

**If we have them I forgot to put it**

assets/templates/tabs-component.html
GusBaamonde: ok, makes sense i check it

PR #307, opened 2020-05-19
fix(Node): cannot use more than 10 version

fcoronelmakingsense:
- Add one more json file to install the dependencies to node version 12 and dont change gulp 3.
 
The explanation is here: https://timonweb.com/posts/how-to-fix-referenceerror-primordials-is-not-defined-error/


GusBaamonde:
❤️

cbernat:
How about adding the version for node and npm in package.json?

PR #312, opened 2020-05-29
fix(Webapp): add styles to first loading

fcoronelmakingsense:
- Add styles to add in webapp the spinner

#Checklist


GusBaamonde:
👍

PR #323, opened 2020-06-26
fix(webapp): header cm logo styles

fcoronelmakingsense:
- fix header cm logo in webapp when the cm user dont have logo and only have the name

image

#Checklist


assets/scss/templates/_header.scss
```diff
@@ -50,6 +49,10 @@
         display: block;
         height: 50px;
       }
+
+      span {
+        line-height: 32px;
```
cbernat, line 54: why not 30 here? to make it a round number

fcoronelmakingsense: changed

cbernat: If 30 is too tight as you've mentioned, leave it that way please.

PR #332, opened 2020-07-20
feat(cards): add the ribbon image

GusBaamonde:
- the pages for the cards and for the doppler plus layout were created
- links to library navigation added
- doppler plus and cards component html created
- four classes were created to modify the position of the ribbons
- the hover effect of the cards was created
- "Data-ribbon" is used to edit the contents of the ribbon

https://cdn.fromdoppler.com/doppler-ui-library/build918/cards-component.html

cards-ribbons

#Checklist

### DIFF
--- a/assets/templates/cards-component.html
+++ b/assets/templates/cards-component.html
@@ -1,0 +1,116 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <title>UI library Doppler - Cards Component</title>
+    <meta name="keywords" content="UI library Doppler - Layout Reports" />
+    <meta name="description" content="First layout of Reports" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link rel="canonical" href="" />
+    <!-- Open Graph metatags -->
+    <meta property="og:title" content="Layout Reports" />
+    <meta property="og:type" content="article" />
+    <meta property="og:image" content="Layout Reports" />
+    <meta property="og:url" content="Layout Reports" />
+    <meta property="og:description" content="Layout Reports" />
+    <meta property="og:site_name" content="Layout Reports" />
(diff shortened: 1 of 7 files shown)


assets/templates/cards-component.html
```diff
@@ -0,0 +1,116 @@
...
+        >
+          Back
+        </a>
+      </header>
+
+      <div class="dp-gray-page p-t-54 p-b-54">
+        <div class="dp-align-center">
+          <div>
+            <h3 class="m-l-30">Card Ribbon left-top</h3>
+            <div class="dp-card-ribbon-lt" data-ribbon="Recomendado"></div>
```
cbernat, line 76: What happens if you add more text here? Say like something that occupies two lines.

cbernat:
image

You need to remove one ../ for the image to be located correctly.

assets/templates/cards-component.html
cbernat: image

This is to think for the future only, that´s all.

cbernat: http://cdn.fromdoppler.com/doppler-ui-library/build917/cards-component.html

GusBaamonde: This does not work like that, there is no space for two words, the design is not prepared for this. It is not an elastic ribbon.
Surely if they want to add more words they will change the design.
This accepts 1 word or 2 very small

GusBaamonde: https://cdn.fromdoppler.com/doppler-ui-library/build918/cards-component.html

cbernat: And in the case we need to support more than one size of ribbon, is it prepared? Or have you thought about this?

GusBaamonde: In all the designs they have there are no variants of ribbons but it is something that can be done without difficulty 🐇

mdirago: Great job Gus! About the ribbons as Gus has said, there aren´t another iterations. Only it´s designed to contain one word, or two short words. If another requerimient shows up we will consider it! :)

PR #333, opened 2020-07-24
feat(card): add images of card

GusBaamonde:
- add styles of the card component
- add the html from card
- add all card states
- adjust the tooltip

See more: https://cdn.fromdoppler.com/doppler-ui-library/build920/cards-component.html

cards-states

#Checklist


assets/scss/modules/_card.scss
```diff
@@ -146,3 +172,144 @@
...
+
+.dpsg-tit-card {
+  font-size: 12px;
+  width: 285px;
+  text-transform: uppercase;
+}
+
+.dp-maximum {
+  border-radius: 3px;
+  background-color: #f2f2f2;
```
cbernat, line 220: Color missing here.

cbernat:
Some images are not showing correctly 
image

https://cdn.fromdoppler.com/doppler-ui-library/build919/cards-component.html

cbernat:
Why is this a ul?

image

I think that without a list would be ok, we don't have more than one price by card right?

cbernat:
The layout is looking very nice! 💪 

image

GusBaamonde:
https://cdn.fromdoppler.com/doppler-ui-library/build920/cards-component.html

GusBaamonde:
> Some images are not showing correctly
> image
> 
> https://cdn.fromdoppler.com/doppler-ui-library/build919/cards-component.html
> 
> Also I don't see the tooltips working 🤔 👀

Resolved! 👍

GusBaamonde:
> Why is this a ul?
> 
> image
> 
> I think that without a list would be ok, we don't have more than one price by card right?

Due to an alignment problem.
The number can grow or the words and everything has to be aligned with the center.
An "ul" solves the whole lineup for us without weird things

mdirago:
Excellent work Gus!

PR #336, opened 2020-08-04
feat: doppler plus general layout created

GusBaamonde:
- doppler plus general layout created
- the free plan was added to the cards
- improved accordion icons for frequently asked questions

https://cdn.fromdoppler.com/doppler-ui-library/build924/doppler-plus.html

layout-doppler-plus2

#Checklist

### DIFF
--- a/assets/scss/modules/_accordion.scss
+++ b/assets/scss/modules/_accordion.scss
@@ -31,16 +31,16 @@
         height: 7px;
         content: '';
         left: 0;
-        top: 29px;
-        border-left: solid 6px $dp-color-grey;
-        border-bottom: solid 6px transparent;
-        border-top: solid 6px transparent;
+        top: 31px;
+        border-left: solid 4px $dp-color-grey;
+        border-bottom: solid 4px transparent;
+        border-top: solid 4px transparent;
         position: absolute;
         transition: all 0.35s;
       }
 
       &.active:before {
-        border-left: solid 6px $dp-color-grey;
+        border-left: solid 4px $dp-color-grey;
         transform: rotate(90deg);
       }
     }
--- a/assets/scss/modules/_card.scss
+++ b/assets/scss/modules/_card.scss
@@ -22,11 +22,16 @@
     border: 2px solid $dp-color-snow;
     text-align: center;
     transition: 0.3s;
-    margin: 0 24px 24px;
+    margin: 0 12px 24px;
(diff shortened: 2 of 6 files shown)


assets/scss/templates/_doppler-plus.scss
```diff
@@ -44,4 +67,35 @@
...
+      color: #333;
+    }
+
+    &:before {
+      width: 5px;
+      height: 7px;
+      content: '';
+      left: 15px;
+      top: 4px;
+      border-left: solid 4px #666;
```
cbernat, line 93: Color hardcoded 😄

cbernat:
When I start to shrink the layout it turns int left aligned like this:
image
👀

GusBaamonde:
> When I start to shrink the layout it turns int left aligned like this:
> 👀

Can be reviewed in another pr

assets/scss/modules/_accordion.scss
```diff
@@ -31,16 +31,16 @@
         height: 7px;
         content: '';
         left: 0;
-        top: 29px;
-        border-left: solid 6px $dp-color-grey;
-        border-bottom: solid 6px transparent;
-        border-top: solid 6px transparent;
+        top: 31px;
```
lucianaDilizio, line 34: What is the reason why is used this kind of numbers? I mean, why not `30px`?

assets/scss/modules/_card.scss
```diff
@@ -54,6 +59,24 @@
       margin: 10px 0 0;
       font-size: 14px;
     }
+
+    .dp-btn-free {
+      justify-content: center;
+      display: flex;
+      align-items: flex-end;
+      height: 170px;
```
cbernat, line 67: Shouldn't all buttons have the same height? or standarized height?

GusBaamonde: This is not a button it only seems like it, but you're right maybe we can push it with another element 👍

assets/scss/modules/_accordion.scss
GusBaamonde: In this case it is a pseudo class for an "icon" in an absolute position, so we can move it to the pixel. In this case we align it to the content. The normalization of spaces that we use in the library is 6px and its multiples, only for paddings and margins. Does not apply in this case.

cbernat:
👍

PR #338, opened 2020-08-07
fix: stakeout of card component

GusBaamonde:
Removal of "data" attributes to manifest a graphical representation, as a bad practice. The "data" attributes are for the manipulation of non-visible data.

docs:
http://microformats.org/wiki/faq#Q:_When_should_I_use_a_microformat.3F_What_are_they_for.3F
https://developer.mozilla.org/es/docs/Learn/HTML/como/Usando_atributos_de_datos#:~:text=Los%20atributos%20data%2D*%20permiten,en%20el%20DOM%20o%20Node.

Look:
https://cdn.fromdoppler.com/doppler-ui-library/build927/cards-component.html

new-ribbon

delete-data-attribute

#Checklist

### DIFF
--- a/assets/templates/cards-component.html
+++ b/assets/templates/cards-component.html
@@ -71,7 +71,7 @@
         <div class="dp-align-center">
           <div>
             <h3 class="dpsg-tit-card m-l-30">Card Ribbon left-top</h3>
-            <div class="dp-card dp-recommended">
+            <div class="dp-card dp-highlighthed">
               <div class="dp-ribbon dp-ribbon-top-left">
                 <span>Nuevo plan</span>
               </div>
@@ -79,7 +79,7 @@
           </div>
           <div>
             <h3 class="dpsg-tit-card m-l-30">Card Ribbon left-bottom</h3>
-            <div class="dp-card dp-recommended">
+            <div class="dp-card dp-highlighthed">
               <div class="dp-ribbon dp-ribbon-bottom-left">
                 <span>Baratija</span>
               </div>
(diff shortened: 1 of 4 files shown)


assets/templates/cards-component.html
```diff
@@ -150,14 +176,21 @@ <h3>Plan name</h3>
...
-              <p>Hasta 100.000 Contactos.</p>
-              <div class="dp-price" data-start="Desde">
-                <div class="dp-amount">
-                  <span class="dp-money-number" data-money="US$">5250</span>
-                  <span>por mes</span>
+            <div class="dp-card-ribbon-rt">
+              <div class="dp-ribbon">
+                <span class="dp-text-ribbon">Recomendado</span>
+              </div>
+              <div class="dp-content-plans">
```
cbernat, line 183: Is it really neccesary to add an extra wrapper div for all extra except the ribbon?

GusBaamonde: I'm working on separating the highlighted state from the ribbon

assets/templates/cards-component.html
```diff
@@ -71,32 +71,54 @@
         <div class="dp-align-center">
           <div>
             <h3 class="dpsg-tit-card m-l-30">Card Ribbon left-top</h3>
-            <div class="dp-card-ribbon-lt" data-ribbon="Recomendado"></div>
+            <div class="dp-card dp-recommended">
```
andresmoschini, line 74: What do you think about _highlighthed_ in place of _recommended_?

```suggestion
            <div class="dp-card dp-highlighthed">
```

PR #343, opened 2020-08-18
fix: fix in tooltip

GusBaamonde:
fix_tootltip

#Checklist

### DIFF
--- a/assets/scss/modules/_tooltip.scss
+++ b/assets/scss/modules/_tooltip.scss
@@ -179,7 +179,7 @@
   .dp-tooltip-container:hover .dp-tooltip-chart,
   .dp-tooltip-container:hover .dp-tooltip-block {
     opacity: 1;
-    z-index: 1;
+    z-index: 1000;
     pointer-events: none;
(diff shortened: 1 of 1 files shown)


assets/scss/modules/_tooltip.scss
```diff
@@ -179,7 +179,7 @@
   .dp-tooltip-container:hover .dp-tooltip-chart,
   .dp-tooltip-container:hover .dp-tooltip-block {
     opacity: 1;
-    z-index: 1;
+    z-index: 1000;
```
cbernat, line 182: Why 1000? could not be a less huge number? like 2 or 3, to keep track of this somehow. Or I'm thinking perhaps you could assign a z index level in a constant for elements, let's say, tooltips: 2, tabs: 1, popups: 3, etc This is too think further

GusBaamonde: You are right, we can lower it to 10. It is not easy to keep track of layers, we know that a modal as a general rule has 9999 maximum layer level, it has to be above everything.
Maybe we can use and jump from 10 to 10.

PR #354, opened 2020-09-10
feat(header-component): apply new styles

fcoronelmakingsense:
- Add new styles and layout to user right menu.

CM: Profile
image

CM: User
image

Demo
image

Enterprise: suscribers
image

Highvolume: emails
image

Standard: suscribers
image

Prepaid
image

Free
image

#Checklist

### DIFF
--- a/assets/scss/templates/_header.scss
+++ b/assets/scss/templates/_header.scss
@@ -207,10 +207,10 @@
       width: 14px;
       padding-top: 1px;
       border-radius: 3px;
-      font-size: 11px;
+      font-size: $dp-sizing--lvl1;
       position: absolute;
       text-align: center;
-      top: 6px;
+      top: $dp-spaces--lv1;
       left: 9px;
       display: none;
     }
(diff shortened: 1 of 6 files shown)


assets/scss/templates/_header.scss
```diff
@@ -368,15 +368,20 @@
...
 
     .user-plan {
       color: $dp-color-green;
       font-size: 11px;
       text-transform: uppercase;
+      margin-top: 10px;
+    }
+
+    .user-plan-enabled {
+      font-size: 11px;
```
cbernat, line 381: I think we have a variable for font sizes...

assets/templates/header-component.html
```diff
@@ -152,18 +152,178 @@
                         ><span class="email">gbaamonde@makingsense.com</span>
                       </p>
                     </header>
-                    <div class="user-plan--container m-t-18">
+                    <!-- <div class="user-plan--container m-t-18">
```
cbernat, line 155: Why do you have all this markup commented? cannot it be left in another page with all examples working?

assets/scss/templates/_header.scss
fcoronelmakingsense: changed

assets/templates/header-component.html
fcoronelmakingsense: changed

PR #358, opened 2020-09-21
fix: modification that the icons are part of the class

GusBaamonde:
**Modification that the icons are part of the class**

iconos-parte-de-clase

#Checklist

### DIFF
--- a/assets/scss/modules/_card.scss
+++ b/assets/scss/modules/_card.scss
@@ -206,9 +206,12 @@
   /// icons card expanded
 
   .dp-icostar {
+    background: url('../img/ico-star.svg') 0 0 no-repeat;
     position: absolute;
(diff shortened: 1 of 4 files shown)


lucianaDilizio (commented):
Nice 😄

assets/scss/modules/_card.scss
```diff
@@ -206,9 +206,12 @@
   /// icons card expanded
 
   .dp-icostar {
+    background: url('../img/ico-star.svg') 0 0 no-repeat;
```
cbernat, line 209: This will need to be fixed in cards screen when this change applies to webapp.

cbernat: I've talked this with Gus. We will fix in webapp.

PR #370, opened 2020-10-16
fix: fix on the backgrounds of the doppler plus

GusBaamonde:
fix on the backgrounds of the doppler plus boxes and add an icon svg for the tabs

icon-and-bgs

#Checklist


lucianaDilizio (commented):
👍

PR #374, opened 2020-10-21
fix: tooltip alignment regardless of size

GusBaamonde:
ico-info-tooltip

https://cdn.fromdoppler.com/doppler-ui-library/build1022/plan-calculator.html

#Checklist


lucianaDilizio (commented):
💪 nice tooltips!

cbernat:
When the resolution is lower, I see the information icon but not the tooltips, perhaps it would be best not to show the i icon when there are no tooltips shown. What do you think @GusBaamonde ?
 
image

PR #376, opened 2020-10-22
fix: elimination of paddings from subscribers' main box

GusBaamonde:
elimination of paddings from subscribers' main box

remove_paddings

nada_raro

#Checklist


lucianaDilizio (commented):
👏

PR #383, opened 2020-10-26
fix in graph of day of week and hours of advanced reports

GusBaamonde:
fix_error_graph

fix_graph

**it's OK now!!**
fix_graph_OK

#Checklist


cbernat:
Great! good catch @GusBaamonde !

PR #388, opened 2020-10-28
fix(UserMenu):  prepaid and monthly plan ui

fcoronelmakingsense:
Before:
image

After:
image


GusBaamonde:
👍

PR #395, opened 2020-11-03
fix: tooltip fix for doppler main lists

GusBaamonde:
Tooltip fix to try to use it in doppler main lists

#Checklist


bcorengia:
Nice! Thanks Gus! This PR prevents the tooltip flickering when hovering it.

PR #399, opened 2020-11-05
DW-565: fix dist version adding csso

cbernat:
### Fix tooltip issue
Dist build and local build were different because csso was not being applied, and that did twist a bit some styles like tooltips styles.

https://cdn.fromdoppler.com/doppler-ui-library/build1055/plan-calculator.html
#Checklist


lucianaDilizio (commented):
💪

PR #411, opened 2020-11-16
DW-539 remove gulp

cbernat:
### Remove gulp from UI library
Replace gulp with npm scripts, to avoid issues with gulp.

- Added watch task for css, html, js
- Added server with watch for development
- Added sourcemaps only for development
- Fixed issue with js minify task (they were two tasks, now we reuse one.)
- Remove gulp
- Remove gulp iconfont and gulp related libraries
- Fixed readme to include new changes

#Checklist


andresmoschini:
👏 👏 👏 

_On change_ looks interesting, I needed something similar in the past, but I did not know how to do it. If it happens again I know who asks about it 😀.

PR #464, opened 2021-02-10
feat: added a container to handle the custom fields and emojis

GusBaamonde:
- added a container to handle the custom fields and emojis
- add the link in the navigation of Big data layouts
- Style sheet for Big data elements was added
- Icons added
- Add a DEMO image at @andresmoschini request 🤪 
- https://cdn.fromdoppler.com/doppler-ui-library/build1106/big-data-input.html

custom-fields-and-emojis2

#Checklist


andresmoschini:
Great! what do you think about using a watermark to be explicit about the image is only for demo purposes?

image

---

@lucianaDilizio please verify if this HTML structure is compatible and friendly with Razor.

assets/templates/big-data-input.html
```diff
@@ -0,0 +1,467 @@
...
+                            <div class="dp-bd-box">
+                              <button
+                                type="button"
+                                class="dp-button button-medium dp-button-bd"
+                                aria-controls="dp-options-cf"
+                              >
+                                <i class="ms-icon icon-brackets"></i>
+                              </button>
+                              <div class="dp-content-bd">
+                                <ul class="dp-list-cfield" id="dp-options-cf">
```
cbernat, line 147: To try an use the js in this library for tabulation you can try:
```suggestion
                                <ul class="dp-list-cfield" id="dp-options-cf" aria-labelledby="cf-button">
```
And then in the button above
```
<button
                                type="button"
                                class="dp-button button-medium dp-button-bd"
                                aria-controls="dp-options-cf"
                                id="cf-button"
                              >
```
Then in javascript index.html try initializing this button:
```
var menubutton = new Menubutton(document.getElementById('cf-button'));
    menubutton.init();
```
Then try and see if tabulation works.

assets/scss/templates/_big-data.scss
```diff
@@ -0,0 +1,94 @@
...
+
+  .dp-bd-box {
+    position: relative;
+  }
+
+  .dp-button-group {
+    position: absolute;
+    right: 0;
+    height: 40px;
+    top: 22px;
```
cbernat, line 13: I have an issue here 
image

For this to work correctly in old doppler I need 60px at top, can you help me with this?

cbernat: Old doppler inputs here in this page are 42px tall, here all styles were overwritten, could you add a class to skip input styles? like `old-input` or perhaps `ignore-library` that allows to have the original styles only. You see here the difference is a lot.  
image

lucianaDilizio: It also breaks the next label and error message tooltip 😕

lucianaDilizio: image
Text format is different too.

cbernat: I think I better take care of this issues in old css, to avoid dirtying the library 😅

assets/templates/big-data-input.html
```diff
@@ -0,0 +1,467 @@
...
+                                </ul>
+                              </div>
+                            </div>
+                            <div class="dp-bd-box">
+                              <button
+                                type="button"
+                                class="dp-button button-medium dp-button-bd"
+                                aria-controls="dp-menu-emojis"
+                              >
+                                <img src="img/emoji.png" />
```
cbernat, line 175: Why do you have the emoji here as png and not as an icon class?

assets/scss/templates/_big-data.scss
cbernat: I can lend you my classes @lucianaDilizio 
image

assets/templates/big-data-input.html
GusBaamonde: I check it with Manu!

cbernat:
@GusBaamonde do you mind if we merge this PR as it is?, and then check keyboard navigation and the emoji icon in another PR? So I can use this in doppler right now pointing to a new version of the library.

PR #482, opened 2021-03-09
feat: add unexpected error and no response

GusBaamonde:
Add unexpected error and no response
Replacement of links by buttons
Validated with design @mdirago

Demo: https://cdn.fromdoppler.com/doppler-ui-library/build1136/subscriber-history.html

DW782-sample

#Checklist


mdirago:
Hi! If it's possible, the link might refresh the specific acordeon row. If it isn't, maybe we can try with a whole page reloading.

bravoy93:
@mdirago yes, it is possible, thanks for the suggestion :)

PR #485, opened 2021-03-11
feat: remove animation on unexpected error from tables

GusBaamonde:
DW790-sample

@mdirago look!
https://cdn.fromdoppler.com/doppler-ui-library/build1142/subscriber-history.html

DW790-sample2

#Checklist

### DIFF
--- a/assets/scss/modules/_tables.scss
+++ b/assets/scss/modules/_tables.scss
@@ -182,7 +182,6 @@
                 position: absolute;
                 left: 40px;
                 top: 26px;
-                animation: flash 2s infinite;
               }
             }
 


andresmoschini:
Are you sure about the blinking icon? 

This reminds me of those good old times of the [FrontPage](https://es.wikipedia.org/wiki/Microsoft_FrontPage).

GusBaamonde:
> Are you sure about the blinking icon?

hahaha we are removing the animation

mdirago:
> Are you sure about the blinking icon?
> 
> This reminds me of those good old times of the [FrontPage](https://es.wikipedia.org/wiki/Microsoft_FrontPage).
> 
> 

We've just checked it with the Design team and ask Gus to remove the blinking animation :)  The reasons that we've discussed were about consistency and that the animation could be distracting from the reload link.

PR #488, opened 2021-03-15
feat: field overlap in dropdown tables

GusBaamonde:
bug-field-overlap

bug-field-overlap2

👀 https://cdn.fromdoppler.com/doppler-ui-library/build1150/subscriber-history.html
@bravoy93 

#Checklist

### DIFF
--- a/assets/scss/modules/_tables.scss
+++ b/assets/scss/modules/_tables.scss
@@ -78,6 +78,7 @@
       .dp-name-text {
         position: relative;
         padding-left: 18px;
+        display: block;
       }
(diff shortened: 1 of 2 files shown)


bravoy93:
Hey @GusBaamonde , take a look at the items in the Consent column and Source Form, I think it can be solved by setting the parent span to `display: flex`
image

bravoy93:
And here it would also be good to add `display: block`
image

bravoy93:
Perfect, it looks great!
image

PR #490, opened 2021-03-19
Dw 716 big data suggestion modal

GusBaamonde:
**Add icons
Add the modal
Add the grid**

👀 : https://cdn.fromdoppler.com/doppler-ui-library/build1165/big-data-input.html

big-data-panel

#Checklist


andresmoschini:
No revisé el código, pero el diseño se ve muy bien y muy adapable!\, creo que es justo lo que necesitamos!

assets/scss/templates/_big-data.scss
```diff
@@ -93,4 +108,157 @@
...
+
+      .advice-disabled {
+        &:before {
+          content: '';
+          position: absolute;
+          left: 0px;
+          top: 4px;
+          width: 20px;
+          height: 20px;
+          background: url('../img/advice-disabled.svg') no-repeat 0 0;
```
cbernat, line 200: Why not add this svg as icon font?

cbernat:
@GusBaamonde could you add an example of Consejos section alone please?

assets/scss/templates/_big-data.scss
GusBaamonde: I wanted to manipulate the entire state with a single class. To use iconfont you had to add one more element and more classes.

cbernat: But to add this into doppler, we need to host the image or change the class. Can you add it as iconfont? If I'm not clear, let´s have a chat.

assets/js/main.js
```diff
@@ -197,3 +197,17 @@ var dopplerUI = dopplerUI || {};
...
   };
+
+  $('.dp-show-tips').on('click', function(){
+    $(this).closest('.field-item').next('.dp-advice').slideToggle();
+
+    if ($(this).find('.icon-close').hasClass('rotation')) {
+        $(this).find('.dp-show-text').html($(this).attr('data-text-hide'));
+      } else {
+        $(this).find('.dp-show-text').html($(this).attr('data-text-show'));
+    }
```
cbernat, line 208: How about this?
```suggestion
    var $showPanelLink = $(this).find('.dp-show-text');
    if ($(this).find('.icon-close').hasClass('rotation')) {
        $showPanelLink.html($(this).attr('data-text-hide'));
      } else {
        $showPanelLink.html($(this).attr('data-text-show'));
    }
```

assets/js/main.js
```diff
@@ -197,3 +197,17 @@ var dopplerUI = dopplerUI || {};
...
+    $(this).find('.icon-close').toggleClass('rotation');
```
cbernat, line 209: Nice! ❤️

assets/scss/templates/_big-data.scss
GusBaamonde: @cbernat I did not understand this point well, this whole panel works under the dp-library class, and the images travel with the library. You have to use the library if or if to put it together. That means that the images are already in the library

cbernat: I've tested this and you are right! It works ok. Thanks for the clarification. If in the future we need this icons in another size perhaps it would be good to move this to icon font, but not now.

cbernat:
Very good job! 🚀

PR #513, opened 2021-04-30
feat: add icons for blocked segments and blocked list

GusBaamonde:
bloqueo

look: https://cdn.fromdoppler.com/doppler-ui-library/build1191/subscriber-history.html

#Checklist


cbernat:
Very nice, but as we've mentioned in the daily, we need this in english and spanish, this is better as an editable text. The idea was very nice though.

assets/img/segments-locked.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="142" height="18" viewBox="0 0 142 18" fill="none" xmlns="http://www.w3.org/2000/svg">
```
cbernat, line 1: This svg is not used anymore right? We should delete this if it's not used anywhere.

GusBaamonde:
icons removed

PR #550, opened 2021-06-23
feat: add pending request status with tooltip in user menu

GusBaamonde:
solicitud-enviada

https://cdn.fromdoppler.com/doppler-ui-library/build1236/nav-user-component.html

#Checklist


cbernat:
This is looking very nice! ❤️

PR #551, opened 2021-06-23
fix: tooltip left class for user menu

GusBaamonde:
tooltip-left-user-menu-01

tooltip-left-user-menu-03

tooltip-left-user-menu-02

#Checklist

### DIFF
--- a/assets/templates/header-component.html
+++ b/assets/templates/header-component.html
@@ -168,6 +168,23 @@
                         </button>
                       </div>
                       <div class="user-plan--type">
+                        <p><strong>498</strong> Suscriptores disponibles</p>
+                        <div class="dp-request-sent">
+                          <button
+                            class="user-plan close-user--menu dp-tooltip-left"
+                            type="button"
+                          >
+                            solicitud enviada
+                            <div class="tooltiptext">
+                              Estamos diseñando un Plan a la medida de tus
+                              necesidades.¡Te contactaremos pronto!
+                            </div>
+                          </button>
+                          <span class="ms-icon icon-info-icon"></span>
+                          <div class="dp-tooltip-container"></div>
+                        </div>
+                      </div>
+                      <div class="user-plan--type">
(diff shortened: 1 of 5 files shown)


assets/templates/header-component.html
```diff
@@ -167,6 +167,27 @@
...
                         </button>
                       </div>
+                      <div class="user-plan--type">
+                        <p><strong>498</strong> Suscriptores disponibles</p>
+                        <div class="dp-request-sent">
+                          <button
+                            class="user-plan close-user--menu"
+                            type="button"
+                          >
+                            solicitud enviada
```
nreal, line 177: This text is only text or label, maybe we need to add css for <p> or <label>, see https://makingsense.atlassian.net/wiki/spaces/DOP/pages/1390051366/Flujo+de+Contrataci+n+de+Plan+Junio+2021

GusBaamonde: I did not understand the question.
I'll call you and we'll see it if you want. 👍

assets/templates/header-component.html
```diff
@@ -167,6 +167,23 @@
                           comprar
                         </button>
                       </div>
+                      <div class="user-plan--type">
+                        <p><strong>498</strong> Suscriptores disponibles</p>
+                        <div class="dp-request-sent">
+                          <button
```
cbernat, line 173: If changes are added into the html take in accound that they should be added into webapp too. 
To validate what I'm asserting please navigate into new reports to check how this dropdown is showing.

GusBaamonde: just add a new item and we check it against previous versions. They are all in the library.
Being a new element, it should not impact what is already there.
I tried the library side in all the cases we use and it works fine
@nreal will have to test how it looks on the screens you mention.

- https://cdn.fromdoppler.com/doppler-ui-library/build1241/nav-user-component.html
- https://cdn.fromdoppler.com/doppler-ui-library/build1241/header-component.html#

PR #590, opened 2021-08-11
fix: adjust information icon size

GusBaamonde:
#Checklist


assets/scss/templates/_doppler-plus.scss
```diff
@@ -208,6 +208,10 @@
   .dp-tabs-plans {
     .tab--link {
       overflow: visible;
+
+      .dp-iconinfo {
+        font-size: 16px;
```
cbernat, line 213: Souldn`t this size also be taken from a variable here? 🤔

PR #591, opened 2021-08-11
DW-1020

GusBaamonde:
Add tag of coming soon for doppler plus

label-coming

#Checklist

### DIFF
--- a/assets/scss/modules/_card.scss
+++ b/assets/scss/modules/_card.scss
@@ -248,4 +248,13 @@
     border-radius: 3px;
     font-size: 10px;
   }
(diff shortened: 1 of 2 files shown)


assets/scss/modules/_card.scss
```diff
@@ -248,4 +248,13 @@
     border-radius: 3px;
     font-size: 10px;
   }
+  .dp-coming {
+    background-color: #fcedec;
+    color: #ec8a83;
+    border: 1px solid #ec8a83;
```
cbernat, line 254: Rogue colors without variables detected 👀

PR #593, opened 2021-08-12
Dw-1004

GusBaamonde:
Enhancement to the flashcard component to include a border property that improves contrast clipping on backgrounds.

border-flashcards

#Checklist

### DIFF
--- a/assets/scss/modules/_messages.scss
+++ b/assets/scss/modules/_messages.scss
@@ -142,8 +142,10 @@
     }
   }
 
+  /* TODO: agregar los colores de los bordes de las flashcards a la paleta de colores de doppler */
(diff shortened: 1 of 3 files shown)


assets/scss/modules/_messages.scss
```diff
@@ -142,8 +142,10 @@
     }
   }
 
+  /* TODO: agregar los colores de los bordes de las flashcards a la paleta de colores de doppler */
   .dp-wrap-success {
     background-color: $dp-color-green-brightness;
+    border: 1px solid #99d6b9;
```
cbernat, line 148: What happened here with the color variables? 👀 `#99d6b9`

PR #608, opened 2021-08-30
fix: build errors

cbernat:
### Fixed two errors on build
- On del-cli module [ERR_REQUIRE_ESM]: Must use import to load ES Module. Fixed using node 12.20.0 I've choosen the version according to [this library debate](https://github.com/manuelbieh/geolib/issues/208) 
- Minifying images '.' is not recognized as an internal or external command --> using npx instead of running the module directly

#Checklist


silsanchez8:
Excelente Ceci!!! 👏  💪  💯

PR #611, opened 2021-09-02
DS-2626-carousel-creation

ornella-688:
### Modifications:
- Slider creation. Html / css / js

image


assets/js/main.js
```diff
@@ -147,75 +147,120 @@ $('.progress-bar').each(function(index) {
...
+
+// Carousel
+
+var carousel = document.querySelector('.dp-carousel');
+var container = carousel.querySelector('.dp-carousel-container');
+var pagination = carousel.querySelector('.dp-carousel-dots');
+var bullets = [].slice.call(carousel.querySelectorAll('.dp-carousel-dot'));
+var totalItems = container.querySelectorAll('.dp-carousel-slide').length;
+var percent = (100 / totalItems);
+var currentIndex = 0;
```
cbernat, line 229: If you have more than one carrousel using this global, won`t work... 🤔

cbernat:
Jenkins build triggered a failure, can you check?
image

assets/js/main.js
```diff
@@ -49,30 +49,30 @@ $(document).on('mouseup', function(e) {
...
-$("#btn-modal-small").click(function() {
+$("#btn-modal-small").click(function () {
   $("#modal-small").fadeIn($delayFade);
 });
-$(".modal .close").click(function() {
+$(".modal .close").click(function () {
   $(".modal").fadeOut();
 });
 
 // This script is ONLY to DEMONSTRATE how to show and hide password, not use in production
```
cbernat, line 74: 💯

cbernat:
image
It adapts very weirdly, the text dissapears 👀
Perhaps it can be done in another PR.

assets/templates/carousel-component.html
```diff
@@ -78,14 +79,77 @@ <h3>Carousel Component</h3>
             <hr class="dpsg-break" />
             <div class="dp-container">
               <div class="dp-rowflex">
-                <div class="col-sm-6 m-b-12">
-                  <div class="dp-box-shadow dpsg-content-sample">
-                    <p class="dp-boxshadow">Carousel 1</p>
+                <div class="col-sm-12 col-md-6">
+                  <div class="dp-carousel dp-carousel-orange">
+                    <div class="dp-carousel-wrapper">
```
cbernat, line 84: Why not add the color in the wrapper? Instead of targeting the wrapper from above.
```suggestion
                    <div class="dp-carousel-wrapper dp-carousel-orange">
```

assets/templates/carousel-component.html
```diff
@@ -78,14 +79,77 @@ <h3>Carousel Component</h3>
...
+                          <span>Último BlogPost</span>
+                          <h3>
+                            ¿Cuáles son las diferencias entre servicio al
+                            cliente y atención al cliente?
+                          </h3>
+                          <a
+                            ><span class="ms-icon icon-arrow-next"></span> Ver
+                            más</a
+                          >
+                        </div>
```
cbernat, line 96: Inside the slide it would be best to differentiate the slide as an articlie with a title and a description, to make it more semantic, in order to make it easier for accesibility, readers and so on.  Something like this:
```suggestion
                        <article class="dp-carousel-slide">
                          <h3>Último BlogPost</h3>
                          <p>
                            ¿Cuáles son las diferencias entre servicio al
                            cliente y atención al cliente?
                            <a
                            ><span class="ms-icon icon-arrow-next"></span> Ver
                            más</a
                          >
                          </p>
                          
                        </article>
```
This way you would read: a slide has a title, with a description, and the description allows you to know more. 
About [article](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article)
Do you see it possible?

assets/templates/carousel-component.html
ornella-688: I thought of it like this in case an element is added outside the wrapper that needs the color format. So I think it is better to add it in the parent element.

assets/templates/carousel-component.html
ornella-688: Yes, in the last commit, the title is being an `h3`, the descriptions a `span` and the "ver más" an `a`. Is it ok this way?

assets/templates/carousel-component.html
cbernat: This is a case of [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it), when we need this we will think about it. 😄

ornella-688: Ok I change it!

assets/js/main.js
```diff
@@ -147,75 +147,97 @@ $('.progress-bar').each(function(index) {
...
+  $(this).find('.icon-close').toggleClass('rotation');
 
-  });
+});
 
+// Carousel
+
+(function () {
+  $(document).ready(function () {
+    return $("input[type='radio']").change(function () {
```
cbernat, line 224: This will affect all apps using the UI library, It would be best like this:
```suggestion
    return $("input[type='radio'][name='carousel1'], input[type='radio'][name='carousel2'] ").change(function () {
```
To target only what we have in the ui library.

ornella-688: I get it. It happens that in that case, if you want to add another carousel, you have to touch the js of the library. Ex, to put another carousel, "carousel3"

assets/js/main.js
```diff
@@ -147,75 +147,97 @@ $('.progress-bar').each(function(index) {
...
+        var $panel, $radiobutton, panel_id;
+        $radiobutton = $(element);
+        panel_id = `#${$radiobutton.attr('id')}_dpslide`;
+        $panel = $(panel_id);
+        if (panel_id === current_panel_id) {
+          return $panel.show();
+        } else {
+          return $panel.hide();
+        }
+      });
```
cbernat, line 239: Why not add the same name for each slide and radios, so you can target them directly?
```
<div class="dp-carousel-slide" id="primero_dpslide" name="carousel1">
  ...
</div>
<div class="dp-carousel-slide" id="segundo_dpslide" name="carousel1">
  ...
</div>
```
So then target al the slides to hide
```suggestion
      var slideId = $(this).prop("id");
      var slidesName = $(this).prop("name");
      // hide all slides
      $('div[name="'+slidesName+'"]').hide();
      // show the one with the correct id
      $(slideId+ '_slide').show();
      });
```
Am I understanding this right?

ornella-688: I tried that and it doesn't work

assets/js/main.js
ornella-688: Since it is calling the slides whose id ends in `_dpslide`, I think it would not affect all the inputs

cbernat: You can try with a console.log, each input type radio will trigger this event. Even if it does nothing but to `_dpslide` ending divs, it is still code running that has not to be in the event. 👀

assets/js/main.js
cbernat: I will try directly on the code, the logic is sound I think 🤔

assets/js/main.js
cbernat: The idea here is that you either use the js in the library or don't.

assets/js/main.js
cbernat:
```
      var slideId = $(this).prop("id");
      var slidesName = $(this).prop("name");
      // hide all slides
      $('div[name="'+slidesName+'"]').hide();
      // show the one with the correct id
      $('#'+slideId+ '_dpslide').show();
```
You can try it this way and it works, I've tested just now. 
Also there is no need to do a return for each event, with just executing the code it's ok. You could replace the whole function by this:
```
// Carousel

  $(document).ready(function () {
    $("input[type='radio'].dp-carousel-dot").change(function () {
      var slideId = $(this).prop("id");
      var slidesName = $(this).prop("name");
      // hide all slides
      $('div[name="'+slidesName+'"]').hide();
      // show the one with the correct id
      $('#'+slideId+ '_dpslide').show();
    });
  });
```

assets/js/main.js
cbernat: This could be replaced by this:
```
 $("input[type='radio'].dp-carousel-dot")
 ```

File: assets/js/main.js (hunk 6 above)
Comment (ornella-688) [reply assets/js/main.js:239]: Great, thanks! I change it

File: assets/scss/modules/_carousel.scss [hunk 7]
```diff
@@ -2,9 +2,193 @@
...
+      }
+
+      .dp-carousel-content .dp-carousel-slide a {
+        color: $dp-color-orange-carousel;
+      }
+
+      .dp-carousel-content .dp-carousel-slide a span {
+        background: $dp-color-orange-carousel;
+        color: $dp-color-white;
+      }
```
cbernat, line 173: This could be a mixin right @GusBaamonde ?

assets/templates/carousel-component.html
```diff
@@ -74,18 +75,157 @@
...
-                    <p class="dp-boxshadow">Carousel 1</p>
+                <div class="col-sm-12 col-md-6">
+                  <div class="dp-carousel">
+                    <div class="dp-carousel-wrapper dp-carousel-orange">
+                      <div class="dp-carousel-content">
+                        <div
+                          class="dp-carousel-slide"
+                          id="primero_dpslide"
+                          name="carousel1"
+                        >
```
cbernat, line 106: You could use a class `active` in the slide that is supposed to show. So then you would have all the slides hidden by styles and change the class to active in the one that must be shown. 
```suggestion
                      <div class="dp-carousel-content">
                        <div
                          class="dp-carousel-slide active"
                          id="primero_dpslide"
                          name="carousel1"
                        >
```
This is more readable in the html, it has more semantic meaning.

cbernat: Of course you will need to add the classes for active and not active:
```
    .dp-carousel-content .dp-carousel-slide{
      display:none;
    }
    .dp-carousel-content .dp-carousel-slide.active{
      display:block;
    }
```

assets/scss/modules/_carousel.scss
GusBaamonde: @ornella-688  this example may help you, you can extend a class and use inheritance that is simpler and semantic:

.boton {
  font-size: 24px;
  padding: .5em .8em;
  border-radius: 3px;
  border: none;
  cursor: pointer;
}
.boton-principal {
  @extend .boton;
  background: #0BE66A;
}
.boton-secundario {
  @extend .boton;
  background: #DDDDDD;
}

assets/scss/modules/_carousel.scss
```diff
@@ -2,9 +2,193 @@
...
+      background: $dp-color-ghostwhite;
+      border-radius: 3px;
+      overflow: hidden;
+    }
+
+    &-content {
+      -webkit-display: inline-flex;
+      -moz-display: inline-flex;
+      -ms-display: inline-flex;
+      -o-display: inline-flex;
```
GusBaamonde, line 27: Remember that all this does not have to be

assets/scss/modules/_carousel.scss
```diff
@@ -2,9 +2,193 @@
...
+      span {
+        font-size: 15px;
+        font-weight: 700;
+        color: $dp-color-lightgrey;
+        margin-bottom: $dp-spaces--lv3;
+        display: block;
+      }
+
+      h3 {
+        font-size: 23px;
```
GusBaamonde, line 76: is there this size?

assets/scss/modules/_carousel.scss
```diff
@@ -2,9 +2,193 @@
...
+      @include media-breakpoint('mobile') {
+        min-height: 300px;
+      }
+
+      &:first-child {
+        display: block;
+      }
+
+      span {
+        font-size: 15px;
```
GusBaamonde, line 68: add variable

assets/scss/modules/_carousel.scss
```diff
@@ -2,9 +2,193 @@
...
+        padding-bottom: $dp-spaces--lv5;
+      }
+
+      a {
+        position: absolute;
+        bottom: 12px;
+        right: 20px;
+        padding-right: $dp-spaces--lv4;
+        display: inherit;
+        font-size: 15px;
```
GusBaamonde, line 89: add variable

assets/scss/modules/_carousel.scss
```diff
@@ -2,9 +2,193 @@
...
+          margin: 0;
+          position: absolute;
+          right: 0;
+          bottom: 0;
+          width: 18px;
+          height: 18px;
+          border-radius: 100%;
+          background: $dp-color-lightgrey;
+          text-align: center;
+          font-size: 8px;
```
GusBaamonde, line 107: is there this size?

assets/templates/carousel-component.html
```diff
@@ -74,18 +75,157 @@
...
+              poner un mismo <b>name</b> para todos los <b>slides</b> del
+              carousel.
+              <br />
+              Ejemplo: para el primer slide de un carousel
+              <b>id="primero_dpslide" name="carousel1"</b>. Para el segundo
+              slide <b>id="segundo_dpslide" name="carousel1"</b>, etc.
+              <br /><br />
+              Poner misma cantidad de inputs que de slides. A cada input ponerle
+              un <b>id</b> y <b>value</b> distinto, y el mismo <b>name</b> que a
+              los slides. <br />
```
GusBaamonde, line 90: the <br> is deprecated

assets/templates/carousel-component.html
```diff
@@ -74,18 +75,157 @@
...
+            <p>
+              Al copiar el html del carousel cambiar <b>id</b> de cada
+              <b>slide</b>, siempre dejando la terminación <b>_dpslide</b>, y
+              poner un mismo <b>name</b> para todos los <b>slides</b> del
+              carousel.
+              <br />
+              Ejemplo: para el primer slide de un carousel
+              <b>id="primero_dpslide" name="carousel1"</b>. Para el segundo
+              slide <b>id="segundo_dpslide" name="carousel1"</b>, etc.
+              <br /><br />
```
GusBaamonde, line 87: the <br> is deprecated

assets/templates/carousel-component.html
```diff
@@ -74,18 +75,157 @@
         <div class="dpsg-content-wrap">
           <article class="dpsg-content">
             <h3>Carousel Component</h3>
-            <p>Nuevo componente de Carousel</p>
+            <p>
+              Al copiar el html del carousel cambiar <b>id</b> de cada
+              <b>slide</b>, siempre dejando la terminación <b>_dpslide</b>, y
+              poner un mismo <b>name</b> para todos los <b>slides</b> del
+              carousel.
+              <br />
```
GusBaamonde, line 83: the <br> is deprecated

assets/templates/carousel-component.html
```diff
@@ -74,18 +75,157 @@
         <div class="dpsg-content-wrap">
           <article class="dpsg-content">
             <h3>Carousel Component</h3>
-            <p>Nuevo componente de Carousel</p>
+            <p>
+              Al copiar el html del carousel cambiar <b>id</b> de cada
+              <b>slide</b>, siempre dejando la terminación <b>_dpslide</b>, y
```
GusBaamonde, line 80: the **b** is deprecated, use **strong**

assets/templates/carousel-component.html
```diff
@@ -74,18 +75,157 @@
...
+              poner un mismo <b>name</b> para todos los <b>slides</b> del
+              carousel.
+              <br />
+              Ejemplo: para el primer slide de un carousel
+              <b>id="primero_dpslide" name="carousel1"</b>. Para el segundo
```
GusBaamonde, line 85: the **b** is deprecated, use **strong**

assets/scss/modules/_carousel.scss
cbernat: I've tried a mixin like this:
```
@mixin colorSlide($primary-color, $border-color, $back-color) {
  border: 2px solid $border-color;
  background: $back-color;

  .dp-carousel-content .dp-carousel-slide span,
  .dp-carousel-content .dp-carousel-slide a {
    color: $primary-color;
  }

  .dp-carousel-content .dp-carousel-slide a span {
    background: $primary-color;
    color: $dp-color-white;
  }
}
```
Then here you should've replace the class content like this
```suggestion
    &-orange {
     @include colorSlide($dp-color-orange-carousel, $dp-color-orange-carousel-border, $dp-color-orange-carousel-back);
```
What do you think @GusBaamonde @ornella-688 ?

ornella-688: Ok, I see it

assets/templates/carousel-component.html
ornella-688: Ok, I do it like this

assets/scss/modules/_carousel.scss
ornella-688: Yes, sorry, I'm deleting it

assets/js/main.js
```diff
@@ -1,8 +1,8 @@
 $('body').on({
-  mouseenter: function() {
+  mouseenter: function () {
     var submenu = $(this).find(".sub-menu");
 
-    if(!$(this).hasClass('submenu-item')) {
+    if (!$(this).hasClass('submenu-item')) {
```
cbernat, line 5: Could you avoid the change in spacing please 🙏  or add this in another commit.  Otherwise when we look for a change in this functionality somewhere in the future, this change will be misleading. There is no real change but tabulation.

cbernat: If I'm not being clear you can ask me directly.

assets/scss/modules/_carousel.scss
ornella-688: No I did it like this on the prototype. But I am changing it with a size of the variables

assets/scss/modules/_carousel.scss
ornella-688: Ok

assets/templates/carousel-component.html
ornella-688: ok, I am deleting it

assets/templates/carousel-component.html
ornella-688: ok, I am changing it

assets/scss/modules/_carousel.scss
cbernat: Don`t use the mixin, I've tried here https://www.sassmeister.com/ and it does add a lot of redundant css 😓 Thanks @GusBaamonde for this tool! It is a better approach to use inheritance.

ornella-688: I think that in this case I can't use inheritance because the styles inside the `orange` and `purple` classes are not the same. It could be in the case that they were the same.

ornella-688: I could only decrease the code like this:

```
   .dp-carousel-orange {
      border-color: $dp-color-orange-carousel-border;
      background: $dp-color-orange-carousel-back;

      .dp-carousel-content .dp-carousel-slide p,
      .dp-carousel-content .dp-carousel-slide a {
        color: $dp-color-orange-carousel;
      }

      .dp-carousel-content .dp-carousel-slide a span {
        background: $dp-color-orange-carousel;
      }
    }
```

assets/js/main.js
ornella-688: ok, no problem

PR #629, opened 2021-09-27
Fix(.gitattributes): add font extensions as binary

silsanchez8:
## Checklist
_* Fix(.gitattributes): add font extensions as binary to fix git extension binaries error._


cbernat:
❤️ 
@ornella-688 will be very happy 😄 I know I will.

GusBaamonde:
❤️

PR #631, opened 2021-09-29
First steps

GusBaamonde:
-Add new steps section
-Add step icons

steps

#Checklist


cbernat:
If we had more than 3 steps we will need add the image for step 4, this will remain like this forever or with the new font then it will be fixed?

cbernat:
image

PR #632, opened 2021-09-30
Details-carousel

ornella-688:
### **Modifications:**
- Added margin bottom
- Slide width increased to 100%

image


ornella-688:
Webapp atomic classes were used and this change was not necessary.

PR #656, opened 2021-11-12
fix: fix in message for user add orange color to button

GusBaamonde:
See:
https://cdn.fromdoppler.com/doppler-ui-library/build1388/account-message-component.html

button-orange

#Checklist


silsanchez8:
😍  Great!

PR #677, opened 2021-12-15
feat: add back button in the new checkout

GusBaamonde:
Add back button in the new checkout
See: https://cdn.fromdoppler.com/doppler-ui-library/build1411/payment-process.html

#Checklist


jhoffman-ms:
🚀

